PR #2, opened 2019-05-08
Edits from PR#1

leem42:


portal-app-template/src/example-configuration/exploreHomeConfiguration/data.tsx
```diff
@@ -34,7 +34,6 @@ export const data: HomeExploreConfig = {
     name: 'QueryWrapperMenu',
     props: {
       rgbIndex,
-      loadingScreen: <div className="bar-loader"><BarLoader color="#47337D" loading={true} /></div>,
```
jay-hodgson: loading?

jay-hodgson: (Michael says not to hardcode the color here)

PR #3, opened 2019-05-10
PR

leem42:
Got tests to work
Got csbc up and running
Renamed directories to be easier to use


app-template/src/config/docTitleConfig.ts
```diff
@@ -0,0 +1,7 @@
+import { DocTitleConfig } from '../types/portal-config'
```
jay-hodgson: remove from checkin

app-template/src/example-configuration/routesConfig.ts
```diff
@@ -0,0 +1,117 @@
+import { GenericRoute } from '../types/portal-config'
```
jay-hodgson: tests could be written to go against real portals

jay-hodgson: but maybe they shouldn't be (can put union of all portal code into example and test that).

configurations/csbc-pson/routesConfig.ts
```diff
@@ -0,0 +1,133 @@
...
+    to: '/',
+    isNested: false,
+    synapseObject: [
+      {
+        name: 'StackedBarChartPreview',
+        title: 'EXPLORE PORTALS',
+        props: {
+          queryWrapperConfigs: [
+            {
+              ...grants.homePageSynapseObject.props
```
jay-hodgson: could simplify (instead of expanding props into a new dictionary that contains the same content)

configurations/csbc-pson/scripts/exportS3ProductionBucketName.sh
```diff
@@ -0,0 +1 @@
+export S3_PRODUCTION_BUCK_LOCATION='TODO'
```
jay-hodgson: do

PR #4, opened 2019-05-13
Added tests

leem42:
Updated README
Updated CSBC PSON
Updated run.sh
Changed StackedBarChartPreview -> StackedBarChartContainer
Added .travis.yml
Updated package.json tests to read from example-configuration
Fixed hover on multi select dropdown

Travis will be broken until there's a patch to SRC that sets the tsconfig module to 'commonJs', otherwise the SRC import will be untranspiled code that react-scripts-ts can't handle.


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
```
leem42: travis will be broken until src gets fixed

leem42: double check

app-template/package.json
```diff
@@ -27,7 +27,9 @@
     "start-js": "react-scripts-ts start",
     "start": "npm-run-all -p watch-css start-js",
     "build": "react-scripts-ts build",
-    "test": "react-scripts-ts test --env=jsdom --setupFiles ./src/setupTests.js",
+    "copy-example-configuration": "cp -r src/example-configuration/ src/config/" ,
+    "run-tests": "react-scripts-ts test --env=jsdom --setupFiles ./src/tests/setupTests.ts" ,
+    "test": "yarn run copy-example-configuration && yarn run run-tests",
```
leem42: put back

app-template/src/App.scss
```diff
@@ -126,29 +132,26 @@ $header-height : 75px;
...
+  @extend %fillerBorder;
 }
 
-.quarter-flex-basis {
-  flex-basis: 50%;
-}
-.nav-button:hover {
-  @extend %hoverFade;
+%bottomBorderNav {
   border-bottom-style: solid;
```
leem42: use filler border with color set

PR #5, opened 2019-05-13
add back yarn build to staging on run.sh

leem42:


run.sh
```diff
@@ -21,33 +21,28 @@ PORTAL_CONFIGURATIONS=configurations
...
+  exit 1
+fi
+chmod +x ./src/config/scripts/exportS3ProductionBucketName.sh
+source  ./src/config/scripts/exportS3ProductionBucketName.sh
+if [ -z "$S3_PRODUCTION_BUCK_LOCATION" ]; then
+  echo 'Error: exportS3ProductionBucketName.sh must export bash variable S3_PRODUCTION_BUCK_LOCATION'
+  exit 1
+fi
+
 if [ "$1" = "WARNING-push-production" ]; then
```
jay-hodgson: not doing a yarn && yarn build if production?

leem42: yep! for the push to production we just sync the contents of the s3 staging bucket with the production bucket

PR #6, opened 2019-05-17
PORTALS-490

jay-hodgson:


app-template/src/AppInitializer.tsx
```diff
@@ -33,6 +36,7 @@ class AppInitializer extends React.Component<AppInitializerProps, AppInitializer
     ).catch((_err) => {
       console.log('no token from cookie could be fetched ', _err)
     })
+    this.updateSynapseCallbackCookie()
   }
 
   componentDidUpdate(prevProps: any) {
```
leem42: Shouldn't updateSynapseCallbackCookie also get called in componentDidUpdate? Otherwise I think it will always redirect back to the first URL put into the browser when it mounts

app-template/src/AppInitializer.tsx
```diff
@@ -49,7 +53,45 @@ class AppInitializer extends React.Component<AppInitializerProps, AppInitializer
...
+    let background = '#4db7ad'
+    let name = ''
+    let icon = ''
+    const footerElement = document.querySelector('#footer')
+    if (footerElement) {
+      color = window.getComputedStyle(footerElement, null).getPropertyValue('color')
+      background = window.getComputedStyle(footerElement, null).getPropertyValue('background-color')
+    }
+    const homeLinkImgElement = document.querySelector('#home-link img')
+    if (homeLinkImgElement) {
```
leem42: a note to myself -- I need to change this to use the footer icon once it gets added

app-template/src/AppInitializer.tsx
```diff
@@ -49,7 +53,45 @@ class AppInitializer extends React.Component<AppInitializerProps, AppInitializer
...
+    const cookieValue = {
+      foregroundColor: color,
+      backgroundColor: background,
+      callbackUrl: window.location.href,
+      logoUrl: icon,
+      portalName: name
+    }
+    this.props.cookies.set(
+      'org.sagebionetworks.security.cookies.portal.config',
+      JSON.stringify(cookieValue), { path: '/', domain: '.synapse.org' })
```
leem42: Should the 'secure' option be provided in the options? https://github.com/reactivestack/cookies/tree/master/packages/react-cookie/#setname-value-options

jay-hodgson: maybe, but not important (not many requests to .synapse.org, and ok to send this silly cookie anywhere)

app-template/src/AppInitializer.tsx
jay-hodgson: Yes

PR #13, opened 2019-05-28
Bumped SRC

leem42:
use putRefreshToken, set isConsistent to True, lazy load on query wrapper


app-template/src/AppInitializer.tsx
```diff
@@ -27,6 +27,17 @@ class AppInitializer extends React.Component<RouteComponentProps & ReactCookiePr
     SynapseClient.getSessionTokenFromCookie().then(
       (sessionToken) => {
         this.setState({ token: sessionToken })
+        if (sessionToken) {
+          SynapseClient.putRefreshSessionToken(sessionToken).then(
```
jay-hodgson: I would expect clearing the session token from the state if this call fails (invalid session token), or maybe only setting the session token in the state if it's still valid.

leem42: updated to only set the session token once its been refreshed and is valid

PR #21, opened 2019-06-03
Added cookies banner

leem42:


app-template/src/CookiesNotification.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  render() {
+    // show banner if they haven't clicked okay, otherwise show nothing
+    return (
+      this.state.hasAgreedToCookies ?
+        false
+        :
+        (
+          <div className="cookiesBanner">
+            This site uses Cookies to enhance your experience and to analyze our traffic. Using Synapse means that you agree with our cookie policy.
```
jay-hodgson: Replace "Synapse" with "this website" for portals (since it might be unclear that this is a Synapse portal)?

PR #22, opened 2019-06-04
PORTALS-486

jay-hodgson:


app-template/sitemap/generate-sitemap.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs')
+fs.readFile('src/config/routesConfig.ts', (err, data) => {
+  if (err) {
+    return console.error(err)
+  }
+  let routesConfigContent = data.toString()
+  //change baseUrl depending on the portal domain
+  const args = process.argv.slice(2)
```
leem42: I think it would help if there was a comment stating that the file expects a domain name as input

jay-hodgson: added missing usage error that talks about the required argument

PR #31, opened 2019-06-14
PORTALS-523

leem42:


app-template/src/portal-components/Layout.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+
+const Layout = ({ children }: { children: any }) => {
+  return (
+    <div className="container-fluid">
+      <div className="row">
+        <div className="col-md-offset-1 col-md-10">
```
jay-hodgson: rethink how to disconnect this layout from SRC

PR #36, opened 2019-06-25
PORTALS-537: create the cookie immediately before following the synapse.org click

jay-hodgson:


app-template/src/AppInitializer.tsx
```diff
@@ -100,7 +111,21 @@ class AppInitializer extends React.Component<RouteComponentProps & ReactCookiePr
    * Will attempt to set a .synapse.org domain cookie that has enough information to lead the user
    * back to this portal after visiting www.synapse.org.
    */
-  updateSynapseCallbackCookie() {
+  updateSynapseCallbackCookie(
```
leem42: If updateSynapseCallbackCookie gets bound in the constructor or is rewritten as `updateSynapseCallbackCookie = (ev: MouseEvent) => {...}` I don't believe you would have to pass in a reference to `this` on [line 67](https://github.com/Sage-Bionetworks/portals/pull/36/files#diff-d59b9d6ae7c75c61234afe55c0711c48R67), and then `window.addEventListener('click', newClickListener)` would be come `window.addEventListener('click', this.updateSynapseCallbackCookie)` and componentWillUnmount() could unbind the callback directly as well instead of through state

jay-hodgson: initially wrote as `updateSynapseCallbackCookie = (ev: MouseEvent) => {...}`, but `this` did not  refer to the component (because this is a global event callback?), which led me to this unfortunate code.

jay-hodgson: If there's a better way to gain access to the component props within this listener, please let me know!

leem42: I believe I ran into this issue before, could you try binding the function in the constructor instead? In [CardFooter](https://github.com/Sage-Bionetworks/Synapse-React-Client/blob/b0dfcf62e3c01b96fc4ae6f55ad0ccb5c34819ac/src/lib/containers/row_renderers/utils/CardFooter.tsx#L91) we have the same setup but with the method bound in the constructor, if that doesn't work then I can pull it in

jay-hodgson: binding did the trick, thanks @leem42 !
cleaned up the code, waiting for successful build...

PR #40, opened 2019-06-26
PORTALS-533

leem42:


jay-hodgson:
nice progress on moving to generic cards.

PR #41, opened 2019-06-28
PORTALS-478

leem42:
Bump SRC and add props to configure project page routing

### DIFF
--- a/app-template/yarn.lock
+++ b/app-template/yarn.lock
@@ -358,8 +358,8 @@
   resolved "https://registry.yarnpkg.com/@types/katex/-/katex-0.5.0.tgz#a7dbb981f1bc89c8b1b5e7fa8a1ce2617ab2f358"
 
 "@types/lodash@^4.14.119":
-  version "4.14.134"
-  resolved "https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.134.tgz#9032b440122db3a2a56200e91191996161dde5b9"
+  version "4.14.135"
+  resolved "https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.135.tgz#d2607c35dd68f70c2b35ba020c667493dedd8447"
 
 "@types/node@*":
   version "12.0.10"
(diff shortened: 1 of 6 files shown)


app-template/yarn.lock
```diff
@@ -8609,6 +8677,12 @@ to-regex-range@^2.1.0:
     is-number "^3.0.0"
     repeat-string "^1.6.1"
 
+to-regex-range@^5.0.1:
```
jay-hodgson: where did all of the new regex helper libraries come from?

leem42: I ran `yarn why to-regex-range` and
got 
```
=> Found "to-regex-range@2.1.1"
info Reasons this module exists
   - "braces#fill-range" depends on it
   - Hoisted from "braces#fill-range#to-regex-range"
info Disk size without dependencies: "16MB"
info Disk size with unique dependencies: "48MB"
info Disk size with transitive dependencies: "88MB"
info Number of shared dependencies: 4
=> Found "sass#to-regex-range@5.0.1"
info Reasons this module exists
   - "sass#chokidar#braces#fill-range" depends on it
   - Hoisted from "sass#chokidar#braces#fill-range#to-regex-range"
info Disk size without dependencies: "16MB"
info Disk size with unique dependencies: "32MB"
info Disk size with transitive dependencies: "72MB"
info Number of shared dependencies: 3
```

it looks like sass got a [patch version bump](https://github.com/Sage-Bionetworks/portals/pull/41/files#diff-639272fd12ecc4bdbe3a16f135ce72abR7867) and these all downloaded as a result

configurations/amp-ad/routesConfig.ts
```diff
@@ -103,10 +103,9 @@ const routes: GenericRoute [] = [
             props: {
               sql: 'SELECT * FROM syn17024229',
               isHeader: true,
-              iconOptions: {},
               type: SynapseConstants.GENERIC_CARD,
               genericCardSchema: {
-                type: 'PROJECT',
+                type: 'Project',
```
jay-hodgson: type could be typed :)

PR #48, opened 2019-07-11
PORTALS-547

leem42:
Overview - 
- Added NF paired down explore pages for each organization
- Changed name of portal component that wraps SRC component with buttons - StatefulButtonControlWrapper and RouteButtonControlWrapper
- The StatefulButtonControlWrapper accepts an array of synapse objects as a result of the paired down explore module
- Also tweaked linkConfig.sh so that when you exit the program it unlinks the symlinked files by default, no more having to run ./linkConfig.sh and ./undoLink.sh
- Bumped SRC


configurations/linkConfig.sh
```diff
@@ -30,4 +42,4 @@ ln -s ../../app-template/src/config/* .
 # start the project
 
 cd ../../app-template/src
-yarn && yarn start
+yarn && yarn start
```
jay-hodgson: warning about newline

app-template/package.json
```diff
@@ -22,7 +22,7 @@
     "react-spinners": "^0.5.3",
     "sass": "^1.5.1",
     "source-map-explorer": "^1.6.0",
-    "synapse-react-client": "1.12.20",
+    "synapse-react-client": "1.12.21",
```
jay-hodgson: This is ancient, we're on 1.12.22! :)

PR #50, opened 2019-07-11
Fix broken csbc staging and links on home page bar chart buttons for amp/csbc

leem42:


configurations/amp-ad/style/_overides.scss
```diff
@@ -12,6 +12,12 @@ footer {
   }
```
leem42: shouldn't need to override the color below like this but its a quick workaround until I fix the css in SRC

PR #51, opened 2019-07-12
PORTALS-545: initialize GA and send event on page location change

jay-hodgson:


app-template/src/AppInitializer.tsx
```diff
@@ -70,7 +70,18 @@ class AppInitializer extends React.Component<RouteComponentProps & ReactCookiePr
   componentDidUpdate(prevProps: any) {
     // https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md
     if (this.props.location !== prevProps.location) {
+      // scroll to the top
       window.scrollTo(0, 0)
+      // send page view event to Google Analytics
```
leem42: If this for some reason this doesn't track analytics as expected then we can use the "react-ga" package as seen in nf [here](https://github.com/Sage-Bionetworks/nf-portal/blob/88031277a843721c8455203a44a3af783e3d159d/nf-portal-site/src/App.js#L22).

jay-hodgson: Sounds good.  I've also toyed around with the idea of creating a separate account ID for each portal (so we would not have to filter the GA View).  Let's see how if the View looks ok after deployment.  If it isn't working as expected, then we may need to make this a parameter in the portal configuration.

jay-hodgson: Thanks @leem42

PR #67, opened 2019-08-16
PORTALS-584: initial checkin for News section

jay-hodgson:


jay-hodgson:
(Wait for small UI cleanup found here: https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/188 , then update SRC dependency)

PR #78, opened 2019-09-03
PORTALS-596: send isInvokingDownloadTable in portal config cookie

jay-hodgson:


jay-hodgson:
Also contains a text fix for PORTALS-643

PR #101, opened 2019-09-23
Tweaks

leem42:
[PORTALS-721](https://sagebionetworks.jira.com/browse/PORTALS-721)
[PORTALS-751](https://sagebionetworks.jira.com/browse/PORTALS-751)
[PORTALS-752](https://sagebionetworks.jira.com/browse/PORTALS-752)
[PORTALS-739](https://sagebionetworks.jira.com/browse/PORTALS-739)


configurations/amp-ad/logoHeaderConfig.ts
```diff
@@ -1,5 +1,5 @@
 import { LogoConfig } from '../types/portal-config'
-import ampAdLogoSvg from './style/amp-ad-logo.svg'
+import ampAdLogoSvg from './style/amp-logo.svg'
```
conundrumgirl: there is actually an svg w/ that name?

leem42: yep! you can see the change here - https://github.com/Sage-Bionetworks/portals/pull/101/files#diff-d7af22d7a81d671b4f31a68c1dca78cd

PR #105, opened 2019-09-25
add robots.txt

jay-hodgson:


jay-hodgson:
@leem42 : besides the missing robots.txt, the domain names are now wrong for the generated amp-ad sitemap.xml (because it assumes it's being hosted on <config>.synapse.org).
How difficult it is to change the amp-ad config to be adknowledgeportal ?

PR #113, opened 2019-09-27
Template pr for updating upload tool props with prod data

leem42:


jay-hodgson:
Surprised I don't see a SRC version bump in here.  Are all of your latest changes to the DrugUploadTool component in SRC 1.12.60?  I would think you would need to bump SRC (to 1.12.61), publish the new version, and then bump it here in portals.

PR #135, opened 2019-10-25
PORTALS-791

leem42:
* PORTALS-791 - added the link config
* Added react-bootstrap to the project
* Refactored navbar to use dropdown from bootstrap, removed extra logic that used bootstrap from css

From SRC version bump:
* PORTALS-646 : added kebab menu to tables
* PORTALS-660 : updated search component
* PORTALS-789 : fixed by unCamelCase function being used
* PORTALS-780  : updated order of keys for tools in amp-ad
* PORTALS-703  : fixed by removing most facet aliases: further discussion is needed for some of the column names, e,g. names with underscores, names that are different altogether, anything that does fit the unCamelCase style


app-template/src/Navbar.tsx
```diff
@@ -157,86 +109,73 @@ export class Navbar extends React.Component<{}, NavbarState> {
...
+                        />
+                        <SvgIcon>
+                          {
+                            // Material expand more svg https://material.io/tools/icons/?icon=expand_more&style=baseline
+                          }
+                          <path d="M16.59 8.59L12 13.17 7.41 8.59 6 10l6 6 6-6z" />
+                        </SvgIcon>
+                  </Dropdown.Toggle>
+                  <Dropdown.Menu className="user-menu-dropdown portal-nav-menu">
+                      <Dropdown.Item
```
conundrumgirl: just wondering shy we are defining the items list explicitly vs creating an array of hrefs and using a map

app-template/src/Navbar.tsx
```diff
@@ -250,37 +189,32 @@ export class Navbar extends React.Component<{}, NavbarState> {
...
+                          {displayName} 
+                        </Dropdown.Toggle>
+                        <Dropdown.Menu className="portal-nav-menu">
+                          {
+                            plainRoutes.map(
+                              (route) => {
+                                if (route.hideRouteFromNavbar) {
+                                  return false
+                                }
+                                const routeDisplayName = route.displayName ? route.displayName : route.name
```
conundrumgirl: could this be simplified to   `const routeDisplayName = route.displayName ||  route.name`

app-template/src/Navbar.tsx
leem42: good point! I hadn't though of that, changed that now

app-template/src/Navbar.tsx
leem42: done

PR #137, opened 2019-10-29
PORTALS-786, PORTALS-686

leem42:
* Bumped SRC - PORTALS-686
* PORTALS-786 - removed facet alias for tumor type, that should fix the issue
* Formatting


configurations/csbc-pson/synapseConfigs/commonProps.ts
```diff
@@ -1,16 +1,15 @@
 export const facetAliases = {
-  'tissue_or_organ.sv': 'Tissue',
-  experimentalStrategy: 'Assay',
-  tumorType: 'Disease',
```
leem42: removing this fixes PORTALS-686

app-template/src/App.scss
```diff
@@ -172,7 +172,7 @@ body {
 .show button.dropdown-toggle.nav-button,
 button.dropdown-toggle.nav-button:hover {
   background: unset !important;
-  color: $primary-action-color;
+  color: unset;
```
leem42: fixes bug with hover states in navbar

conundrumgirl: Didn't know 'unset' existed. Cool!

PR #139, opened 2019-10-30
PORTALS-852

leem42:
* PORTALS-852 - added class to home page explore object on each portal except drug tool (doesn't have the component) which adds a gray background and padding
* Formatting from linter

### DIFF
--- a/app-template/src/App.scss
+++ b/app-template/src/App.scss
@@ -381,6 +381,11 @@ div.SRC-marginBottomTen {
   }
 }
 
+.home-explore-container {
+  background: #f5f5f5;
+  padding-bottom: 60px;
+}
+
 #exploreCount {
   padding-top: 20px;
 }
--- a/app-template/src/portal-components/Layout.tsx
+++ b/app-template/src/portal-components/Layout.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 const Layout = ({
   children,
-  containerClassName,
+  containerClassName = '',
 }: {
   children: any
   containerClassName?: string
--- a/app-template/src/portal-components/Layout.tsx
+++ b/app-template/src/portal-components/Layout.tsx
@@ -1,12 +1,18 @@
 import * as React from 'react'
 
-const Layout = ({ children }: { children: any }) => {
+const Layout = ({
+  children,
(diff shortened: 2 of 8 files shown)


app-template/src/types/portal-config.d.ts
```diff
@@ -16,88 +16,93 @@ declare type CardContainerLogic = {
...
 }
 
 type GenerateComponentsFromRow = {
-  name: 'GenerateComponentsFromRow',
+  name: 'GenerateComponentsFromRow'
   props: GenerateComponentsFromRowProps
 }
 
-type Metatdata = {
+type Metadata = {
```
leem42: fixed type here...

app-template/src/types/portal-config.d.ts
```diff
@@ -16,88 +16,93 @@ declare type CardContainerLogic = {
...
   title?: string
   link?: string
   style?: React.CSSProperties
   isOutsideContainer?: boolean
+  // applied to the inner most container of the object
   className?: string
+  // applied to outer most container of the object
+  containerClassName?: string
```
leem42: new container class

configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -17,15 +30,31 @@ const routes: GenericRoute [] = [
       {
         name: 'StatefulButtonControlWrapper',
         title: 'EXPLORE CONTENT',
+        containerClassName: 'home-explore-container',
```
leem42: using new prop to add class to home page object

app-template/src/App.scss
```diff
@@ -381,6 +381,11 @@ div.SRC-marginBottomTen {
   }
 }
```
leem42: corresponding class for the home page explore module

app-template/src/portal-components/Layout.tsx
```diff
@@ -1,12 +1,18 @@
 import * as React from 'react'
 
-const Layout = ({ children }: { children: any }) => {
+const Layout = ({
+  children,
+  containerClassName = '',
```
leem42: adding class here to outer most container of the synapse object

PR #157, opened 2019-12-18
fix issues with types being dropped

leem42:
This has been an ongoing issue for sometime where the types weren't being respected by typescript. Fixed by importing all the component props from SRC explicitly in the portal-config file. This isn't completely fixed for all the portal components whose types aren't being registered.

There are a few TODOs to remove the ts-ignore comments, once SRC is bumped and pulled into this those won't be needed.


configurations/adknowledgeportal/synapseConfigs/studies.ts
```diff
@@ -2,14 +2,22 @@ import { HomeExploreConfig, SynapseConfig } from '../../types/portal-config'
...
+  CardLink,
+} from 'synapse-react-client/dist/containers/CardContainerLogic'
 import studyHeaderSvg from '../style/study-header.svg'
 
 const unitDescription = 'studies'
 const rgbIndex = 0
 export const studiesSql = 'SELECT * FROM syn17083367'
 const sql = studiesSql
 const facet = 'Species'
+const cardLink: CardLink = {
```
leem42: remove

PR #158, opened 2019-12-19
PORTALS-927

leem42:
- Added entityId prop
- Refactored card props in a few places to reuse same variable
- Changed SRC type imports to use new folder name
- Also updated prop for linking in the title

### DIFF
--- a/app-template/src/types/portal-config.d.ts
+++ b/app-template/src/types/portal-config.d.ts
@@ -115,10 +115,6 @@ export type SynapseConfig = (
 ) &
   Metadata
 export type SynapseConfigArray = SynapseConfig[]
-type P = SynapseConfig['props']
-const x: P = {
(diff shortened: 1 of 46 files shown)


app-template/src/types/portal-config.d.ts
```diff
@@ -56,12 +59,12 @@ type Markdown = {
 
 type StatefulButtonControl = {
   name: 'StatefulButtonControlWrapper'
-  props: StatefulButtonControlProps
+  props: any
```
conundrumgirl: any way we can do better than 'any'?

leem42: There was an odd issue with the typing that I was running into here, I'd like to leave it as is and I can add a TODO comment in another PR because I don't think it will be an easy bug to solve, thanks for pointing this out though.

PR #166, opened 2020-01-08
PORTALS-968

leem42:
Currently there are two folders in the repo: app-template and configurations which distinguish between the live code and the configuration code. The issue with this approach is -
- requires symlinking app-template/types to configurations, which doesn't always work
- there are two typescript configurations, .e.g tsconfig.json/package.json, when there only needs to be one
- symlinking the configuration to the app-template can result in deleting code

This pr -
- Removes configurations and app-template folders so that all the code lives in the same directory, which illuminated a few issues with the types
- Adds logic to the symlinking process that notifies when there was an error and a way to fix this
- Allows relative paths throughout the project


.gitignore
```diff
@@ -1,6 +1,6 @@
 node_modules/
 build/
-app-template/src/config/**
+src/config/
```
leem42: config -> target/dist/build

linkConfig.sh
```diff
@@ -0,0 +1,81 @@
...
+folderNoSlash=${1%/}
+
+CONFIG_FOLDER=src/config
+ACTIVE_CONFIGURATION=src/configurations/$folderNoSlash
+
+# we want to always make sure that the linking is undone after this script so we capture
+# ctrl-c. This has removed various bugs when developing locally.
+# See here - https://unix.stackexchange.com/a/407249
+trap handleInt SIGINT
+function handleInt {
```
leem42: undoLink

src/types/portal-config.d.ts
```diff
@@ -128,17 +127,15 @@ export interface BaseRoute {
...
-  synapseConfigArray: SynapseConfigArray
+  synapseConfigArray?: SynapseConfigArray
 }
 
 export interface NestedRoute extends BaseRoute {
   isNested: true
   routes: Array<BaseRoute | NestedRoute>
-  synapseConfigArray?: SynapseConfigArray
-  to?: string
   icon?: string
```
leem42: icon not needed

PR #173, opened 2020-01-23
PORTALS-744

leem42:
Take all stateful logic from navbar and put in app initializer, added callback for when new synapse client login/sign out functionality is added


src/Navbar.tsx
```diff
@@ -76,41 +54,22 @@ export class Navbar extends React.Component<{}, NavbarState> {
...
-            userProfile,
-            token: newToken,
-          })
-        })
-        .catch(_err => {
-          console.error('user profile could not be fetched ', _err)
-        })
-    }
+
+  goToTop = (_event: any) => {
```
conundrumgirl: do we need _event here?

src/AppInitializer.tsx
```diff
@@ -33,14 +43,14 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
...
-      .querySelector('meta[name="description"]')!
-      .setAttribute('content', docTitleConfig.description)
+  resetSession = () => {
+    SynapseClient.signOut(this.getSession)
+    this.setState({
+      showLoginDialog: false,
+    })
+  }
 
+  getSession = () => {
```
conundrumgirl: Should this return a promise?

src/AppInitializer.tsx
```diff
@@ -221,15 +257,11 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
...
-      {
-        path: '/',
-        domain: domainValue,
-        expires: expireDate,
-      },
-    )
+    this.props.cookies.set(COOKIE_CONFIG_KEY, JSON.stringify(cookieValue), {
+      path: '/',
+      domain: domainValue,
+      maxAge: 10,
```
leem42: double check maxage

src/AppInitializer.tsx
leem42: I just made a commit to flatten out the function with async/await and also consolidated the logic to be easier to read, thanks for pointing that out! 

I couldn't figure out how to make this function return as a promise though. Since this function is called by synapseclient, I'd like to avoid synapseclient working any harder by calling this function as a promise, but I did add a helper function initAnonymousUserState which I think better pieces out this function.

PR #174, opened 2020-01-27
PORTALS-775: add (hidden) anchor for each route to the dom

jay-hodgson:


jay-hodgson:
Screen Shot 2020-01-27 at 10 48 03 AM

PR #175, opened 2020-01-29
PORTALS-349: expose SRC version only

jay-hodgson:
since this will catch the majority of changes


jay-hodgson:
Screen Shot 2020-01-28 at 4 12 32 PM

PR #183, opened 2020-02-10
PORTALS-775: use standard router instead of hash-based

jay-hodgson:
TODO
- update SRC
- deploy to a staging portal bucket.  
- update staging portal bucket settings, and cloudfront distribution (this will eventually need to be done on 7 other buckets, if successful)

### DIFF
--- a/src/Navbar.tsx
+++ b/src/Navbar.tsx
@@ -196,12 +196,11 @@ class Navbar extends React.Component {
                         }
                         const routeDisplayName =
                           route.displayName || route.name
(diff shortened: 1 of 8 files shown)


src/Navbar.tsx
```diff
@@ -226,7 +224,7 @@ class Navbar extends React.Component {
                               <Dropdown.Item
                                 key={route.name}
                                 className="SRC-primary-background-color-hover SRC-nested-color center-content"
-                                href={`#${route.to}`}
+                                href={`${route.to}`}
```
jay-hodgson: no longer need `

PR #191, opened 2020-02-24
add detail pages to support deep linking on explore page

leem42:


jay-hodgson:
Please change other portals as well (to support deep linking in the future)

PR #195, opened 2020-02-27
PORTALS-1045

leem42:


src/portal-components/csbc-home-page/AboutPortal.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type AboutPortalProps = {
+  title: string
+  token?: string
+  cardProps: HomePageCardProps[]
+  logos: string[]
+  ownerId?: string
+  wikiId?: string
+}
+
+const AboutPortal = (props: AboutPortalProps) => {
```
jay-hodgson: I wonder if this will be reused in other portals!  If so, we should refactor into SRC, it's a nice well-written component.

PR #246, opened 2020-04-20
PORTALS-1188: force page reload on back/forward navigation

jay-hodgson:
read in current query params and reflects in app


leem42:
We could optimize this one step further by checking if the window.location.pathname _hasn't_ changed onpopstate.

If you navigate from explore/projects to explore/datasets the app will automatically load the different routes no problem. You only want it to reload when it goes from explore/projects?param=**xyz** to explore/projects?param=**abc**, pathname is the same but the url params have changed.

jay-hodgson:
> We could optimize this one step further by checking if the window.location.pathname _hasn't_ changed onpopstate.
> 
> If you navigate from explore/projects to explore/datasets the app will automatically load the different routes no problem. You only want it to reload when it goes from explore/projects?param=**xyz** to explore/projects?param=**abc**, pathname is the same but the url params have changed.

Agree that your suggestion is preferable, but I'm not sure what hooks to use to be informed of the url changes (in all cases on all browsers).  There are [some ideas about redefining history.pushState](https://stackoverflow.com/questions/4570093/how-to-get-notified-about-changes-of-the-history-via-history-pushstate) (and presumably replaceState) to add this hook, but I'm not confident in that solution.

PR #254, opened 2020-04-22
link datasets and tool to publication detail pages, link to pubmed on…

leem42:
… detail page, fix link click from home page to explore page


src/configurations/csbc-pson/synapseConfigs/onPointClick.ts
```diff
@@ -10,6 +49,11 @@ export const onPointClick = ({
...
     facet = 'consortium'
   }
+  const encodedQuery = generateEncodedQueryForURL(
+    typeUpperCase,
+    facet,
+    facetValue,
+  )
   // @ts-ignore
-  window.location = `/Explore/${typeUpperCase}?facet=${facet}&facetValue=${facetValue}`
+  window.location = `/Explore/${typeUpperCase}?QueryWrapper0=${encodedQuery}`
```
conundrumgirl, line 58: this will work if there is a single QueryWrapper on the page or if there is more than one it would work with the first.

PR #277, opened 2020-05-20
CRC researcher portal

jay-hodgson:


src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react'
```
leem42, line 1: can remove this import and make this file .ts instead, sometimes typescript decides unused imports should break the build (and other times it doesn't :) )

src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { PartialRow, TableUpdateTransactionRequest } from 'synapse-react-client/dist/utils/synapseTypes/Table/TableUpdate'
+import { Row } from 'synapse-react-client/dist/utils/synapseTypes'
+import { SynapseClient } from 'synapse-react-client'
+
+const handleParticipantWorkflowChange = async (event:any, newWorkflowState:string) => {
```
leem42, line 6: this event:any is not ideal, should really use a type here but I see that we forgot to export the CustomControlCallbackData in SRC, shoot.

leem42: You could duplicate the type in this file
CustomControlCallbackData = {
  data: QueryResultBundle | undefined
  selectedRowIndices: number[] | undefined
  refresh: () => void
}
typescript will see the typing of the event for the function and cross reference when you use it in routeConfig

src/configurations/crc-researcher/routeButtonControlWrapperProps.ts
```diff
@@ -0,0 +1,16 @@
+import { SynapseConfig } from 'types/portal-config'
+import { RouteButtonControlWrapperProps } from 'portal-components/RouteButtonControlWrapper'
+const routeButtonControlProps: RouteButtonControlWrapperProps = {
+  // this has to get overriden,
+  synapseConfig: {} as SynapseConfig,
+  colors: ['#47337D', '#407BA0', '#7798AC', '#77BBBF', '#58A158', '#3C4A63'],
+  customRoutes: [],
```
leem42, line 7: Shouldn't customRoutes have names for each route to render the buttons on the /Participants page? 

It should be
['Uncategorized', 'Selected', 'Invited','Hidden']

Otherwise you'll only navigate through the tabs at the top

src/configurations/crc-researcher/synapseConfigs/hidden.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const hidden: SynapseConfig = {
+  name: 'QueryWrapperPlotNav',
+  props: {
+    rgbIndex,
+    entityId,
+    shouldDeepLink: true,
+    sql,
+    name: 'Hidden Participants',
+    // @ts-ignore
```
leem42, line 19: shouldn't need the ts-ignore anymore

src/configurations/crc-researcher/handleParticipantWorkflowChange.tsx
```diff
@@ -0,0 +1,45 @@
...
+          entityId,
+          toAppend: {
+            concreteType: 'org.sagebionetworks.repo.model.table.PartialRowSet',
+            tableId: entityId,
+            rows: rowUpdates
+          }
+        }]
+      }
+      const token = await SynapseClient.getSessionTokenFromCookie()
+      SynapseClient.updateTable(request, token).then(() => {
```
leem42, line 39: Having a .catch seem's like a good idea, there isn't any recourse with the .catch but it really helps with debugging later.

PR #298, opened 2020-06-02
PORTALS-1362: use svg in header, absolute position text on header

leem42:
background-image has some quirky features regarding svg sizing, see here http://whereswalden.com/2011/10/21/properly-resizing-vector-image-backgrounds/, its not the same sizing that a standard svg uses. Changed portal config to use an actual SVG if defined in the header to gain more control of the sizing.

I believe the reason the icon for digitalhealth is so difficult is because the pattern in the svg actually matters when it scales. The design artwork for the other portals doesn't show any particular object, so this wasn't an issue.

One of the biggest issues that I was running into is that the svg's intrinsic size didn't seem to get respected by the background-size property, so the height wasn't being respected, I tried different values background-size/height and to no avail. The best I could get was having an image with the wrong aspect ratio. Using the img html element I also ran into the same issue.

Before this PR:
image

After this PR:
image


src/Header.tsx
```diff
@@ -2,9 +2,17 @@ import * as React from 'react'
...
+    summary,
+    title,
+    showBlur = true,
+    centerText = false,
+    HeaderSvg,
+  } = headerConfig
+  const hasImg = HeaderSvg !== undefined
   return (
     <header id="header">
+      {hasImg && <HeaderSvg />}
```
emmanmills, line 15: This looks good to me. I agree that if we want to scale the image properly without cropping, adding the image directly to the markup is easier.

src/configurations/digitalhealth/style/_overides.scss
```diff
@@ -1,9 +1,23 @@
...
+  .container-fluid {
+    position: absolute;
+    top: 20%;
+    width: 80%;
+  }
+}
+
 #footer {
-  background-blend-mode: multiply;
+  background-color: #3e7293 !important;
```
emmanmills, line 20: It's great to remove the blend mode css. I can't think of a time to use it. Do you think we can apply the new background color without using !important? Like making the selector more specific might just do the trick. I am worried that !important is often harder for others to override this selector later on if the footer needs to be customized somehow.

leem42: Removed the important tag, good call!

PR #312, opened 2020-06-12
fix spelling homepage

kelshmo:
Two small fixes to the banner on the homepage. 

As I am unfamiliar with this code base this may not be the right fix - I'm interested in helping where I can.


leem42:
thanks for the PR!

PR #329, opened 2020-06-30
crc design changes.  rename selected to potential, add scheduled and …

jay-hodgson:
…tested (selecting columns that do not yet exist)


jay-hodgson:
CRC researcher portal changes only

PR #331, opened 2020-07-01
CRC issue 593: add an overview participant stacked bar plot.

jay-hodgson:
fed by aggregate query against participant table, using WorkflowState

### DIFF
--- a/src/portal-components/crc-researcher/ParticipantsBarPlot.tsx
+++ b/src/portal-components/crc-researcher/ParticipantsBarPlot.tsx
@@ -1,0 +1,182 @@
+import React, { FunctionComponent, useEffect, useState } from 'react'
+import Plotly from 'plotly.js-basic-dist'
+import * as PlotlyTyped from 'plotly.js'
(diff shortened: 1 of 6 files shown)


jay-hodgson:
For more information, please see the associated issue: https://github.com/Sage-Bionetworks/covidrecoverycorps/issues/593

src/portal-components/crc-researcher/ParticipantsBarPlot.tsx
```diff
@@ -0,0 +1,182 @@
...
+      setIsLoaded(true)
+    })
+    .catch(err => {
+      throw err
+    })
+  return () => {}
+}, [token])
+
+  return (
+    <>
```
leem42, line 164: Does the component need to send a message if the user isn't logged in?

jay-hodgson: It's a good point, but I'd rather not tie this UI to each component.  
I don't want to see n components on the page all reporting that you need to sign in (lessons learned from synapse.org).

PR #332, opened 2020-07-06
add inviteSentOn and dataGroups to select. removed unreferenced imports

jay-hodgson:


jay-hodgson:
Changes to crc-researcher only.

PR #351, opened 2020-07-23
update metadata query

kelshmo:
I made a [mistake with the schema](https://sagebionetworks.jira.com/wiki/spaces/PEC/pages/844759171/PEC+Studies+Schema) (now updated). We want to also include `dataSubtype` and `assay` variables.


leem42:
Looks good to me, thanks for the PR!

PR #362, opened 2020-08-10
PORTALS-1479: Use Here-Document

xschildw:
[PORTALS-1479]

[PORTALS-1479]: https://sagebionetworks.jira.com/browse/PORTALS-1479


jay-hodgson:
If the other readers are like me, you'll need [more information](https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Redirections)
It looks like you could use <<- to support natural tab indentation, but that's just a nitpick.  Looks great!

PR #369, opened 2020-08-18
Change GenerateComponentsFromRow -> DetailsPage

leem42:
I believe that DetailsPage is a more intuitive name than GenerateComponentsFromRow


jay-hodgson:
I like this change, no need to expose the implementation details in the component name (describing what it is, rather than how it works).

PR #380, opened 2020-08-25
Home page css tweaks

leem42:
added alternating home page background and css
image

bump src - more css tweaks to the people's module

### DIFF
--- a/src/_App.scss
+++ b/src/_App.scss
@@ -1,3 +1,4 @@
+$primary-bgcolor-rgba: unset; // need to remove in src
 @import 'config/style/_overides.scss';
 @import 'node_modules/synapse-react-client/dist/template_style/Index.scss';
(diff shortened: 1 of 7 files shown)


src/_App.scss
```diff
@@ -17,12 +18,26 @@
   }
 }
 
+// TODO: Make this class generalizable
+.home {
+  padding-bottom: 50px;
+  &:nth-child(odd) {
```
emmanmills, line 24: I wonder what happen if not the 'odd' children have the color background, but instead the 'even' children like this design: https://www.figma.com/file/KiY0iNeez6ZkqdqP35SLS6/Homepage-Modules?node-id=1519%3A322. We might also have cases in the future when the background color only apply to some even or odd children. What do you think?

emmanmills: Also if we later add additional components on the home page, the odd/even order might get messed up.

PR #390, opened 2020-09-01
PORTALS-1575

leem42:
- restructure psychencode to contain top level file `resources.ts` with table identifiers, allowing portal owners to more easily version the portal by update just that file
- Bump src for bug fixes: fix sql parser encountering versioned synId, fix dataset cards for NF, news feed updates


jay-hodgson:
This could be a nice pattern to adopt for all portal configurations.  
@emmanmills , could you review?

PR #402, opened 2020-09-03
Wait to Merge: Portals-1519 add mobile topnav

emmanmills:


src/Navbar.tsx
```diff
@@ -16,7 +17,7 @@ type SynapseSettingLink = {
   settingSubPath?: string
 }
 
-class Navbar extends React.Component {
+class Navbar extends React.Component<any, any> {
```
leem42, line 20: could we use a type here for the state instead of `any`, e.g.
```
type State = {
  showMenu: boolean
}
class Navbar extends React.Component<any, State> {...
```

src/_App.scss
```diff
@@ -229,6 +247,18 @@ button.dropdown-toggle.nav-button:hover {
   @extend %bottomBorderNav;
 }
 
+// Adjustment for desktop version after adding mobile topnav bar
+.nav-link-container button.dropdown-toggle,
+.nav-link-container button.dropdown-toggle:hover {
+  background: none;
+  color: #404b63;
```
leem42, line 254: should this color be $primary-action-color?

src/_App.scss
```diff
@@ -212,6 +212,24 @@ body {
...
+.top-nav-button, a.top-nav-button {
+  text-transform: uppercase;
+  font-weight: bold;
+  height: -webkit-fill-available;
+  // fallback for mozilla and ie11
+  height: 100%;
+  font-size: 16px;
+  text-align: center;
+  justify-content: center;
+  color: #404b63;
```
leem42, line 224: same question as below

src/Navbar.tsx
emmanmills: Good idea! Code updated.

src/_App.scss
emmanmills: This is for text color, which is dark grey. I see it hardcoded in other places in _App.scss so I just follow it.

src/_App.scss
emmanmills: Same answer as above.

PR #414, opened 2020-09-08
after talking to Lynn, we should only show participants who have completed all surveys

jay-hodgson:
(in the workflow step tables)


jay-hodgson:
and not show a plot-based facet nav for allSurveysComplete

PR #417, opened 2020-09-09
PORTALS-1519: Remove nav-button class

emmanmills:
Remove nav-button class that causes the layout issue. 

Screen Shot 2020-09-09 at 11 48 44 AM

### DIFF
--- a/src/Navbar.tsx
+++ b/src/Navbar.tsx
@@ -313,7 +313,7 @@ class Navbar extends React.Component<any, State> {
                   return (
                     <NavLink
                       key={topLevelTo}
-                      className={`top-nav-button nav-button nav-button-container center-content ${this.getBorder(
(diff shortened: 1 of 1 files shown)


src/Navbar.tsx
```diff
@@ -313,7 +313,7 @@ class Navbar extends React.Component<any, State> {
                   return (
                     <NavLink
                       key={topLevelTo}
-                      className={`top-nav-button nav-button nav-button-container center-content ${this.getBorder(
+                      className={`top-nav-button nav-button-container center-content ${this.getBorder(
```
leem42, line 316: I thought we might need the nav-button class for desktop view, does the site look okay without it?

emmanmills: Yes, the top-nav-button has the desktop css also.

PR #424, opened 2020-09-11
PORTALS-1633: parser does not parse new HAS clause properly.

jay-hodgson:
for now use old LIKE (matches amp-ad)


emmanmills (approved):
Looks good to me.

PR #431, opened 2020-09-14
Portals-1519: Remove user account links from mobile view

emmanmills:
Base on design feedback: https://sagebionetworks.jira.com/browse/PORTALS-1519?focusedCommentId=120306

1. Move user icon and user account link to a new file so it can be used for both mobile and desktop.
2. Remove user account links from mobile view, instead, view account link will go directly to Synapse's user account page.
3. Add a sign out button at the end of the dropdown menu in mobile view. 

Screen Shot 2020-09-14 at 2 36 48 PM


src/portal-components/NavUserLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return(
+    <>
+      <UserCard
+        userProfile={userProfile}
+        size={SynapseConstants.SMALL_USER_CARD}
+        preSignedURL={userProfile?.clientPreSignedURL}
+        hideText={true}
+        link="javascript:void(0)"
+      />
+      <SvgIcon className="arrow-down">
```
jay-hodgson, line 24: Consider pulling in material ui icons dependency for definition.  Or, if that's too heavy-handed, refactor this arrow-down icon into it's own component for future reuse.

jay-hodgson (approved):

src/portal-components/NavUserLink.tsx
emmanmills: I have updated the code to pull the icon directly from the material ui icons package.

PR #432, opened 2020-09-15
PORTALS-960: remove all of these custom loading screens.  

jay-hodgson:
also delete the htan config (abandoned)


emmanmills (approved):
Looks good!

PR #433, opened 2020-09-15
PORTALS-1525: use RssFeedCards pointing to new WP site.

jay-hodgson:
Support top level link route (News nav bar item points to external site).


src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -452,14 +452,15 @@ const routes: GenericRoute[] = [
     ],
   },
   {
-    to: 'News',
     isNested: false,
-    synapseConfigArray: news,
+    displayName: 'News',
+    to: undefined,
```
emmanmills, line 457: Question. Do we need to pass "undefined" here otherwise the route won't work?

emmanmills (approved):
Looks good and approved. Just 1 question for clarification.

src/configurations/adknowledgeportal/routesConfig.ts
jay-hodgson: I think technically it will work (because of the effective `el.link ?? el.to` in NavBar), but it makes sense to set this to `undefined` if we're using the link (and not using the react router).

PR #437, opened 2020-09-17
PORTALS-1664: add custom icons to organization page.

jay-hodgson:


jay-hodgson:
Need to update SRC, publish a new version, and pull that into here before merging this PR.

package.json
```diff
@@ -16,7 +16,6 @@
     "clone-deep": "^4.0.1",
     "node-sass": "4.12.0",
     "npm-run-all": "^4.1.5",
-    "plotly.js-basic-dist": "^1.47.3",
```
jay-hodgson, line 19: deleted duplicate (plotly.js-basic-dist import is also listed on line 32/33)

src/configurations/nf/organizations.ts
```diff
@@ -81,9 +82,10 @@ export const organizationDetailsPage: SynapseConfig[] = [
         title: 'organizationName',
         type: SynapseConstants.ORGANIZATION,
         description: 'summary',
-        icon: 'logo',
+        icon: 'abbreviation',
```
jay-hodgson, line 85: column used for mapping

package.json
emmanmills: sounds good

emmanmills (approved):

PR #457, opened 2020-10-01
Portals 1675 new study details page

emmanmills:
1. Add config settings to turn on tab layout for Psychencode
2. Re-arrange details page content to different tabs based on design
3. Add Explore 2.0 in "Explore Data" tab

Screen Shot 2020-10-01 at 12 12 36 PM


package.json
```diff
@@ -7,7 +7,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.4",
     "@fortawesome/free-solid-svg-icons": "^5.3.1",
     "@fortawesome/react-fontawesome": "^0.1.3",
-    "@material-ui/core": "^3.9.3",
+    "@material-ui/core": "^4.11.0",
```
emmanmills, line 10: I upgraded the material ui core because I was going to use their tab component. But later decided not to because it will require a lot of overriding the material ui styling.

src/configurations/psychencode/synapseConfigs/studies.ts
```diff
@@ -135,21 +158,27 @@ export const details: DetailsPageProps = {
...
+      name: 'Markdown',
       props: {
-        sql: dataSql,
-        rgbIndex,
-        title: 'Data',
-        sqlOperator: 'HAS',
+        ownerId: 'syn4921369',
+        wikiId: '477467',
       },
+      title: 'Access Requirements',
```
emmanmills, line 169: Move Access Requirement after Metadata

src/portal-components/DetailsPage.tsx
```diff
@@ -18,19 +18,25 @@ import { SynapseConfig } from 'types/portal-config'
...
 import injectPropsIntoConfig from './injectPropsIntoConfig'
-import { cloneDeep } from 'lodash'
 import { ExternalFileHandleLink } from 'synapse-react-client/dist/containers/ExternalFileHandleLink'
 import { BarLoader } from 'react-spinners'
+import SvgIcon from "@material-ui/icons/ExploreOutlined"
+import ExploreOutlinedIcon from '@material-ui/icons/ExploreOutlined'
+import {LockedFacet} from "synapse-react-client/dist/containers/QueryWrapper"
+import {ReactComponent as StudySVG} from "synapse-react-client/dist/assets/icons/study2.svg"
+import {ReactComponent as DataSummarySVG} from "synapse-react-client/dist/assets/icons/Data2.svg"
+
```
emmanmills, line 28: I have to direct import the svg. These icons are not available from Material UI. And our ICON component uses image tag  . The image tag prevents me from customizing the fill colors of the svg icons.

src/portal-components/DetailsPage.tsx
```diff
@@ -309,8 +330,56 @@ export default class DetailsPage extends React.Component<
...
   }
 
-  renderSynapseConfigArray = () => {
+  renderTabs() {
+    const { tabIndex } = this.state
+    return(
+      <div className="tab-groups">
+        <span className={tabIndex === 0 ? "tab-item-active" : "tab-item"} onClick={(e) => {
+          this.setState({tabIndex: 0})
+        }}><StudySVG></StudySVG>Study Details</span>
```
jay-hodgson, line 339: This is currently hard-coded to support a Study details page.  An enhancement would be to include the tabs definition as part of the config.

For example, instead of a hard-coded icon, this could use an icon type/value string and IconOptions that are used elsewhere (then this might become <Icon iconOptions={iconOptions} value={iconValue} type={type} /> for every tab config).

jay-hodgson: And if icon fill color is presenting an additional complexity, let's not customize the color for now.

src/portal-components/DetailsPage.tsx
```diff
@@ -418,6 +494,13 @@ export default class DetailsPage extends React.Component<
       const injectedProps = injectPropsIntoConfig(value, el, {
         ...deepCloneOfProps,
       })
+
+      // For explorer 2.0, cannot assign key `lockedFacet` to deepCloneOfProps due to type errors,
+      // assign lockedFacet value directly to injectedProps only if resolveSynId.value is true
+      if (resolveSynId && resolveSynId.value) {
```
jay-hodgson, line 500: I think you could use the null propagation operator to make this just:
resolveSynId?.value

emmanmills: This is cool! And it is fixed. Thank you!

src/portal-components/DetailsPage.tsx
```diff
@@ -240,25 +244,41 @@ export default class DetailsPage extends React.Component<
...
-        <div className="DetailsPage">
-          <div className="button-container">{this.renderMenu()}</div>
+        <div className="DetailsPage tab-layout">
           <div className="component-container" ref={this.ref}>
-            {synapseConfigContent}
+            {<DetailsPageTabs
+              tabConfigs={tabLayout}
+              loading={isLoading}
+              tabContents={!isLoading && this.buildTabContent()}>
+            </DetailsPageTabs>}
```
emmanmills, line 256: Here I moved the tab code to its own file. Tab settings are from synapse portals configs. Tab contents are passed as an array to the DetailsPageTabs component.

jay-hodgson (approved):
Looks great, thanks Emma!  I think AMP-AD is going to want to define a tabbed details page too.

PR #473, opened 2020-10-14
PORTALS-1212: Change study details page to 2-tab layout

emmanmills:
Screen Shot 2020-10-14 at 1 48 32 PM

### DIFF
--- a/src/configurations/adknowledgeportal/synapseConfigs/studies.ts
+++ b/src/configurations/adknowledgeportal/synapseConfigs/studies.ts
@@ -181,10 +181,9 @@ export const studiesDetailsPageProps: DetailsPageProps = {
       name: "QueryWrapperPlotNav",
       tabIndex: 1,
       props: {
-        rgbIndex: 1,
+        rgbIndex,
         name: 'Data',
         visibleColumnCount: 10,
-        // sqlOperator: 'HAS',
         tableConfiguration: {
           showAccessColumn: true,
           columnLinks: [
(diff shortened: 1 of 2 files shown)


src/configurations/adknowledgeportal/synapseConfigs/studies.ts
```diff
@@ -137,43 +174,36 @@ export const studiesDetailsPageProps: DetailsPageProps = {
...
-      title: 'Related Studies',
-      tableSqlKeys: ['Study'],
-      props: {
-        sqlOperator: '=',
-        sql: studiesSql,
-        ...studyCardConfiguration,
+      resolveSynId: {
+        value: true,
       },
+      tableSqlKeys: ['study'],
```
emmanmills, line 205: What is this tableSqlKeys for? Can I remove it?

jay-hodgson: I think this tells the details page what key value should be injected, working hand in hand with resolveSynId.  So on this study page, the url params have study=syn5550404 for example, but this component will get the resolved value (the name of syn5550404 which is 'MSBB') which will be used to filter.

emmanmills: Nevermind, I just found out that it needs to be there.

src/portal-components/DetailsPage.tsx
```diff
@@ -448,7 +448,8 @@ export default class DetailsPage extends React.Component<
           value = name
 
           // For explorer 2.0, construct an object to contain the locked facet name and facet value
-          lockedFacet.facet = columnName
+          // Make sure locked facet name is lowercase 
+          lockedFacet.facet = columnName.toLowerCase()
```
jay-hodgson, line 452: Not sure if this is always true, but understand why this might be necessary in the AD Knowledge portal ("Study" vs the file view "study" column).  Will work in PsychEncode today as well ("study" is already lowercase across tables).

jay-hodgson (approved):

src/configurations/adknowledgeportal/synapseConfigs/studies.ts
emmanmills: Thank you, that makes sense!

src/portal-components/DetailsPage.tsx
emmanmills: Yes, the lowercase function is added because AD Knowledge portal uses "Study" so it doesn't match the file view "study" column when comparing the 2 strings, therefore the locked facet was not working.

PR #507, opened 2020-11-11
Test versioned view with download list in PsychENCODE portal

kelshmo:
Let's get this versioned view into staging to see how it works!

This PR includes the versioned view that drives the Explore Data and Download list component and includes a fix to the Grants data table.


kelshmo:
@jay-hodgson would it be appropriate to modify my permissions so I can request reviewers?

PR #508, opened 2020-11-13
Portals 1749:  Update NF study and organization details pages

emmanmills:


src/configurations/nf/organizations.ts
```diff
@@ -23,49 +34,68 @@ export const organizationDetailsPageConfig: DetailsPageProps = {
...
+        searchConfiguration: {
+          searchable: [
+            'datasetName',
+            'summary',
+            'studyName',
+            'diseaseFocus',
+            'manifestation',
+            'fundingAgency',
+          ],
+        },
```
emmanmills, line 95: The `searchConfiguration` is copied from the study details page settings. Do we need this `searchConfiguration` in organization/funder details page?

src/portal-components/DetailsPage.tsx
```diff
@@ -428,7 +428,12 @@ export default class DetailsPage extends React.Component<
     return split.map((splitString) => {
       let value = splitString.trim()
       let entityTitle = ''
-      let lockedFacet: LockedFacet = {}
+
+      // For explorer 2.0, construct an object to contain the locked facet name and facet value
+      const lockedFacet: LockedFacet = {
+        facet: columnName
+      }
+
```
emmanmills, line 436: Unlike Psychencode or AD knowledge portals, NF has resolveSynid set equal to false, we need to set the locked facet to column name regardless there is a resolveSynid object.

src/configurations/nf/organizations.ts
jay-hodgson: Likely yes (to define the search columns that make sense).  Ideally this would be defined once, and reused in all of these different places (so in the future when it needs to be modified, we don't have to hunt them all down!)

src/configurations/nf/style/_overides.scss
```diff
@@ -12,3 +12,12 @@ $primary-action-color: #404B63;
   background-color: #125E81;
   background-blend-mode: multiply;
 }
+
+// Customized for nf.
+// Per design, the table should extend to full page width but there aren't enough columns to fill the empty space.
+.DetailsPage.tab-layout .QueryWrapperPlotNav{
+  table.grip-flex {
+    width: 100% !important;
```
jay-hodgson, line 20: Is this a general change? (Can this be applied to all QueryWrapperPlotNav table.grip-flex ?)

src/portal-components/DetailsPage.tsx
jay-hodgson: ok

src/configurations/nf/organizations.ts
emmanmills: Good idea! I moved the search config settings to `commonProps.ts``

src/configurations/nf/style/_overides.scss
emmanmills: Yes, I moved the fix to SRC: https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/765

jay-hodgson (approved):

PR #518, opened 2020-12-08
PORTALS-1763

jay-hodgson:


src/Navbar.tsx
```diff
@@ -294,7 +295,7 @@ class Navbar extends React.Component<any, State> {
                         <Dropdown className={this.getBorder(topLevelTo)}>
                           <Dropdown.Toggle
                             variant="light"
-                            id={displayName}
+                            id={`Navbar-dropdown-${displayName}`}
```
jay-hodgson, line 298: This was "Organizations" in one portal, which matched a title.  The original solution (that used react-router-hash-link) relied on the element being scrolled to having an ID that matched the hash value (just like standard html).  That's no longer the case, but I left in this change.

src/Navbar.tsx
```diff
@@ -345,14 +346,32 @@ class Navbar extends React.Component<any, State> {
...
               routesConfig.filter((el) => !el.hideRouteFromNavbar).length <
                 7 && (
-                <NavLink
-                  key={'Home'}
-                  className={`top-nav-button nav-button-container center-content ${isHomeSelectedCssClassName} ${this.getBorder(
-                    '',
-                  )}`}
-                  to={'/'}
-                  text={'Home'}
-                />
```
jay-hodgson, line 355: Change the Home button so that it's now a Dropdown.

src/Navbar.tsx
```diff
@@ -345,14 +346,32 @@ class Navbar extends React.Component<any, State> {
...
+                <Dropdown className={this.getBorder('')}>
+                  <Dropdown.Toggle
+                    variant="light"
+                    id={'Navbar-dropdown-Home'}
+                    className={`nav-button-container top-nav-button ${isHomeSelectedCssClassName}`}                    
+                  >
+                    {/* Clicking the Home item immediately brings you to the Home page (and opens the dropdown, if there are items) */}
+                    <NavLink to='/' text='Home' />
```
jay-hodgson, line 356: Clicking on the Home dropdown button still takes you to the home page (but also pops open the dropdown menu, so you can jump to the section that you're interested in).

src/RouteResolver.tsx
```diff
@@ -116,7 +117,13 @@ const RouteResolver: React.FunctionComponent<RouteComponentProps> = ({
   if (document.title !== newTitle) {
     document.title = newTitle
   }
-
+  const scrollToRef = React.useRef<HTMLElement>(null)
```
jay-hodgson, line 120: If the hash value (from the URL) is set, and it matches the title of one of the Synapse components on the page, then it will scroll to it (with an offset to respect the top nav menu height).

src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -30,13 +30,16 @@ const routes: GenericRoute[] = [
       },
```
jay-hodgson, line 30: Some of the components in CancerComplexity had their titles embedded in their custom components.  Now it sets the title in the standard way (to support the Home menu dropdown, and make them all consistent)

src/configurations/nf/routesConfig.ts
```diff
@@ -41,37 +41,55 @@ const routes: GenericRoute[] = [
         },
```
jay-hodgson, line 41: Moved a couple of components around in the nf config, to be closer to the design.

src/AppInitializer.tsx
```diff
@@ -101,7 +101,7 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
     // But it's best practice to clean up the global listener on component unmount.
     window.addEventListener('click', this.updateSynapseCallbackCookie)
     // on first time, also check for the SSO code
-    SynapseClient.detectSSOCode()
+    SynapseClient.detectSSOCode()
```
emmanmills, line 104: Looks like this is just adding blank spaces. Would you like to revert the change?

src/RouteResolver.tsx
```diff
@@ -129,6 +136,7 @@ const RouteResolver: React.FunctionComponent<RouteComponentProps> = ({
           subtitle,
           props,
         } = el
+        const scrollToJsx = title && hash && hash == `#${encodeURI(title)}` ? <span ref={scrollToRef} /> : <></>
```
emmanmills, line 139: Do you think we should use strict equal here? 
hash == ``#${encodeURI(title)}`

emmanmills (approved):
Looks good! Thanks.

PR #531, opened 2020-12-17
PORTALS-1792: Move upset plot data sql to resources file

emmanmills:


nickgros (changes requested):
The entity is also hard-coded here: https://github.com/Sage-Bionetworks/portals/blob/4f13647d53d24975187496caee3897afce71fe6e/src/configurations/psychencode/synapseConfigs/data.ts#L5

I don't think that variable is used anywhere, but can you verify that and remove it, otherwise override it in resources.ts?

emmanmills:
@nickgros , I removed the unused constant.

nickgros (approved):

PR #540, opened 2021-01-12
new about page wiki

kelshmo:
My goal is to keep the content the same but point to a wiki sub-page, instead of the Project wiki main page.


jay-hodgson:
Sorry @kelshmo , I just noticed this old PR.  Is it still valid (what you want in psychencode)?

kelshmo:
yes!

PR #541, opened 2021-01-14
PORTALS-1593

nickgros:
https://sagebionetworks.jira.com/browse/PORTALS-1593?filter=-1

In this PR:
* SCSS has been refactored to take advantage of Dart Sass modules. The most significant change is splitting `_overides.scss` to `_variable_overrides.scss` and `_style_overrides.scss`
  * All non-configuration SCSS files now refer to a `_variables.scss` file to get the primary/secondary colors and any other variables.
  * `_variable_overrides.scss` will override the values in `_variables.scss` because we specify the configuration to have a higher priority in the SCSS load order.
    * We must have a `_variable_overrides.scss` file outside of the configuration, set at a lower priority in the load order, to successfully compile the SCSS. 
  * `_style_overrides.scss` contains custom CSS used in each Portal. This is loaded at the end of `App.scss`. This was split from the variable overrides because the variable overrides must be loaded first, which would cause many custom rules to be overridden by the default rule with identical specificity.
    * Similarly, we must have a `_style_overrides.scss` file outside of the configuration, set at a lower priority in the load order, to successfully compile the SCSS. 
* Use new secondary action colors defined in PORTALS-1593
  * If design/stakeholders do not want the secondary color in a particular portal, we can easily change the secondary color to match the primary color
* Change colors from primary to secondary where identified in the mock-up (see images in the JIRA)
* Remove `$primary-text-color`. This is unused in both portals and SRC.


package.json
```diff
@@ -39,8 +39,8 @@
     ]
   },
   "scripts": {
-    "build-css": "sass src/:src/ --load-path=.",
-    "watch-css": "yarn run build-css && sass src/ --watch --load-path=.",
+    "build-css": "sass src/:src/ --load-path=src/config/style --load-path=src/style --load-path=.",
```
nickgros, line 42: The load-path order here is crucial; if `src/config/style` is not first, then the `_variable`/`_style_overrides.scss` files outside of the portal configuration (which are required to compile SCSS without a configuration) will take precedence.

src/App.scss
```diff
@@ -1,8 +1,19 @@
-$primary-bgcolor-rgba: unset; // TODO: need to remove in src
-@import 'config/style/_overides.scss';
-@import 'node_modules/synapse-react-client/dist/template_style/Index.scss';
-@import './portal-components/csbc-home-page/style/all.scss';
-@import 'node_modules/synapse-react-client/dist/style/main.scss';
+@use 'variable_overrides';
+@use 'variables' as Portal;
```
nickgros, line 2: We load the variable overrides (highest priority based on load order). Then we load the variables themselves, which have been overridden by the overrides.

src/App.scss
```diff
@@ -562,3 +576,5 @@ div.SRC-marginBottomTen {
     }
   }
 }
+
+@import 'style_overrides';
```
nickgros, line 580: We load the highest-priority `_style_overrides.scss` at the end to ensure they override selectors with the same specificity.

src/Markdown.scss
```diff
@@ -1,4 +1,5 @@
-@import 'config/style/overides';
+@use 'variable_overrides';
```
nickgros, line 1: Not strictly necessary, but I think we should load the overrides in every top-level file that uses variables, because I'm not sure if it's defined behavior for the Sass compiler to have been smart enough to already load it.

src/portal-components/csbc-home-page/style/_ecosystem.scss
```diff
@@ -30,9 +34,12 @@
         padding: 10px;
         &.isSelected,
         &:hover {
-          color: $primary-action-color;
-          background-color: rgba(71, 51, 125, 0.1);
```
nickgros, line 34: Changed this to a transparent variant of the primary color. We do this in a similar component in SRC.

src/App.scss
```diff
@@ -1,8 +1,19 @@
...
-@import 'node_modules/synapse-react-client/dist/template_style/Index.scss';
-@import './portal-components/csbc-home-page/style/all.scss';
-@import 'node_modules/synapse-react-client/dist/style/main.scss';
+@use 'variable_overrides';
+@use 'variables' as Portal;
+
+@use 'node_modules/synapse-react-client/dist/style/main.scss' with (
+  $primary-action-color: Portal.$primary-action-color, 
+  $secondary-action-color: Portal.$secondary-action-color,
+  $primary-bgcolor-rgba: unset,
```
jay-hodgson, line 7: main primary-bgcolor-rgba not overridden by variables, but instead always unset?

src/App.scss
emmanmills: Does the order matter here? I am just curious because 'variable_overrides' is supposed to override 'variables' right?

src/portal-components/csbc-home-page/style/_ecosystem.scss
emmanmills: This is great!

src/README.md
```diff
@@ -40,7 +40,7 @@ test-configuration/
 │   ├── exportS3ProductionBucketName.sh     
 │   └── exportS3StagingBucketName.sh        
 └── style                                   
-    ├── _overides.scss                      # contains main theme colors
+    ├── _overrides.scss                      # contains main theme colors
```
jay-hodgson, line 43: This document seems inaccurate.  I guess I expected the test-configuration to have the same changes as other portals configurations (delete _overides.scss, and create _style_overrides.scss and _variable_overrides).

jay-hodgson: README could be expanded to explain the scss compilation strategy.

src/App.scss
emmanmills: While this is not essential, I want to point out that this could be a problem if someone forgets or doesn't know the loading order of css and add a bunch of css after your import statement. I wonder if you could create a new file with just 2 imports to prevent that issue.
```
@import 'App';
@import 'style_overrides';
```

Or App.scss has 2 imports
```
@import '<new file that contains App's css ';
@import 'style_overrides';
```

src/README.md
nickgros: Yup--I updated this once, made more changes, and then forgot to update it again. Will fix + add content to describe the whole process

src/App.scss
nickgros: Before this PR, the first line in this file was
```scss
$primary-bgcolor-rgba: unset; // TODO: need to remove in src
```

We don't currently set a specific value for this in any of the portals. Here, I am maintaining whatever the current behavior is, as I haven't looked into what's actually happening here.

src/App.scss
nickgros: I _think_ that the Sass compiler is smart enough to determine if a file is being overridden regardless of the order. Ultimately, it shouldn't matter, because `@use` is designed to load each file only once (this contrasts with `@import`, where a file will be loaded each time it is called). 

In my mind, putting the overrides before the variables makes more sense because the overrides are actually setting the values in the 'variables' file, rather than overwriting a global variable.

I am curious to see what happens with something like:

```scss
@use 'variables' as x;
@use 'overrides1';

div.a {
  background-color: x.$color;
}

@use 'overrides2'

div.b {
  background-color: x.$color;
}

```

When would `x.$color` take on the value of `$color` in `overrides1.scss` or `overrides2.scss`?

src/App.scss
nickgros: I think this is a good idea. Will make this change.

src/_Markdown.scss
```diff
@@ -1,4 +1,3 @@
-@use 'variable_overrides';
```
nickgros: Removed these because these are now imported by App.scss.

src/App.scss
emmanmills: Got it now! I'd just make sure the default css in `'variables' as Portal` doesn't overwrite the ones in `variable_overrides`.

jay-hodgson (approved):

emmanmills (approved):
Looks great! Thank you.

PR #546, opened 2021-01-22
PORTALS-1783: Enable portal card icon for nf

emmanmills:


src/configurations/nf/synapseConfigs/studies.ts
```diff
@@ -54,6 +54,21 @@ export const studyCardConfiguration: CardConfiguration = {
     baseURL: 'Explore/Studies/DetailsPage',
     URLColumnName: 'studyId',
     matchColumnName: 'studyId',
+  },
+  columnIconOptions: {
+    iconBasePath: "/icons/",
```
jay-hodgson, line 59: May want to pull this out as a constant (make things easier if we ever want to move things around).

jay-hodgson (approved):

src/configurations/nf/synapseConfigs/studies.ts
emmanmills: I added a file `src/utils/PortalsConstants.ts` to save all the constants for portals.

PR #550, opened 2021-01-28
PORTALS-1803: change AD metadata table to use explore 2.0, but hide visualizations by default.

jay-hodgson:
Also filter the facets (for left-hand facet value navigation).


emmanmills (approved):
Looks good!

PR #554, opened 2021-02-03
PORTALS-1497:  If resulting component definition returns a react fragment only, don't try to wrap with a title

jay-hodgson:


emmanmills (approved):
Looks good!

PR #561, opened 2021-02-11
PORTALS-1811: update AD program logo

jay-hodgson:


amapeters:
@jay-hodgson we had a request from the NIA to change the acronym for this program from 'ELP' to 'ELITE'. It looks like it broke the link to the logo

amapeters:
https://adknowledgeportal.synapse.org/

PR #568, opened 2021-02-23
PORTALS-1851: use portal header logo (instead of footer logo)

jay-hodgson:
and callback banner background will be white

### DIFF
--- a/src/AppInitializer.tsx
+++ b/src/AppInitializer.tsx
@@ -200,16 +200,8 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
         }
       }
     }
-    let color = '#4db7ad'
-    let background = 'white'
     let name = ''
     let icon = ''
-    const footerElement = document.querySelector('#footer')
-    if (footerElement) {
-      color = window
-        .getComputedStyle(footerElement, null)
-        .getPropertyValue('background-color')
-    }
     const logoImgElement = document.querySelector('#header-logo-image')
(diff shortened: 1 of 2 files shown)


jay-hodgson:
Screen Shot 2021-02-22 at 4 35 12 PM

jay-hodgson:
Related to https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4520

src/AppInitializer.tsx
```diff
@@ -200,22 +200,19 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
...
     let icon = ''
     const footerElement = document.querySelector('#footer')
     if (footerElement) {
       color = window
-        .getComputedStyle(footerElement, null)
-        .getPropertyValue('color')
-      background = window
         .getComputedStyle(footerElement, null)
         .getPropertyValue('background-color')
     }
```
emmanmills, line 212: You delete the background statement and not the color state in `if (footerElement)` block. Is this intentional?

emmanmills (commented):
Just a comment to make sure I am not missing anything. Otherwise looks good.

src/AppInitializer.tsx
jay-hodgson: Yep, it now uses the footer background color as the themed color in the back-bar.  And the back-bar background color is just white now.

emmanmills (approved):

PR #582, opened 2021-03-12
Portals-1847: Fix Mobile Top Nav

emmanmills:
While looking at the responsive spacing issue for the Explore Bar I found some layout issue on the mobile top nav. Checked with design, the current explorer bar spacing looks good as is. No change is needed. But I did fix the mobile top nav in this jira.

### DIFF
--- a/src/_Core.scss
+++ b/src/_Core.scss
@@ -195,7 +195,7 @@ body {
     box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.5);
     background: rgba(256, 256, 256, 0.9);
     .nav-button {
-      color: #1A1C29;
+      color: Portal.$dark-gray;
     }
   }
   &.explore-nav {
@@ -203,7 +203,7 @@ body {
     .nav-button {
       color: #515359;
       &.isSelected {
-        color: #1A1C29;
+        color: Portal.$dark-gray;
       }
     }
   }
--- a/src/_Core.scss
+++ b/src/_Core.scss
@@ -229,8 +229,8 @@ body {
   text-align: center;
   justify-content: center;
   text-decoration: none;
-  padding: 0px 25px !important;
-  margin: 0px !important;
+  padding: 0px 25px;
+  margin: 0px;
   min-width: 100px;
   white-space: normal;
   @extend %fillerBorder;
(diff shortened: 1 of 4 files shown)


src/_Core.scss
```diff
@@ -229,8 +229,8 @@ body {
   text-align: center;
   justify-content: center;
   text-decoration: none;
-  padding: 0px 25px !important;
-  margin: 0px !important;
```
emmanmills, line 233: @nickgros , do you know why we need the !important here? I deleted them but if it's going to break something in Portals, please let me know. Thanks.

src/_Core.scss
```diff
@@ -272,7 +272,7 @@ a.nav-button {
 // override src behavior
 .show button.dropdown-toggle.nav-button,
 button.dropdown-toggle.nav-button:hover {
-  background: unset !important;
```
emmanmills, line 275: @nickgros , do you know why we need the !important here? I deleted them but if it's going to break something in Portals, please let me know. Thanks.

src/_Core.scss
nickgros: I have no idea why it's used. It looks like the .nav-button class is only used in the navbar, so if your changes look good, then I think you can safely remove them

nickgros: I have no idea why it was originally there. Seems like these styles apply to all of the `nav-button`s just fine without `!important`, so I think it's safe to remove.

src/_Core.scss
nickgros: Same here, I don't know why it was there, but it seems like it shouldn't be !important, especially since that breaks the mobile view.

src/_Navbar.scss
```diff
@@ -120,11 +127,14 @@
       }
       .dropdown-menu > li > a {
         padding: 3px $mb-menu-item-padding;
+        color: #1A1C29;
```
nickgros, line 130: If this is a color we would want to use in other places, consider a variable.

nickgros (approved):

src/_Navbar.scss
emmanmills: Good idea!

PR #586, opened 2021-04-13
PORTALS-1883: use pubMedId to join publications with datasets and tools

jay-hodgson:


src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -274,27 +273,27 @@ const routes: GenericRoute[] = [
...
                     {
                       name: 'CardContainerLogic',
-                      columnName: 'publicationTitle',
+                      columnName: 'pubMedId',
                       title: 'Related Tools',
-                      tableSqlKeys: ['publicationTitle'],
+                      tableSqlKeys: ['pubMedId'],
                       props: {
-                        sqlOperator: 'LIKE',
+                        sqlOperator: 'HAS',
```
vpchung, line 296:
```suggestion
                        sqlOperator: '=',
```

vpchung (changes requested):
The new Tools table we're using has the `pubMedId` column as a STRING rather than a STRINGLIST, so we'll have to update sql operator from `HAS` to `=`.  Other than that, everything works locally on my end!

PR #608, opened 2021-05-12
PORTALS-1924: Add experimental mode button in portals footer

emmanmills:
Screen Shot 2021-05-12 at 11 08 15 AM

### DIFF
--- a/src/Footer.tsx
+++ b/src/Footer.tsx
@@ -3,6 +3,7 @@ import footerConfig from './config/footerConfig'
 import logoFooterConfig from './config/logoFooterConfig'
 import { ReactComponent as PoweredBySvg } from './portal-assets/poweredbysynapse.svg'
 import Versions from 'portal-components/Versions'
+import ExperimentalMode from "synapse-react-client/dist/containers/ExperimentalMode";  // synapse-react-client/dist/utils/functions/sqlFunctions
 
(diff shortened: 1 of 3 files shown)


src/Footer.tsx
```diff
@@ -30,6 +31,9 @@ export const Footer: React.SFC<{}> = () => {
         <div id="footer-logo-link">{logo}</div>
       </div>
       <div id="portal-contact-footer" className="center-content">
+        <span className={"bootstrap-4-backport experimental-mode"}>
+          <ExperimentalMode />
```
jay-hodgson, line 35: Looking at this in the context of a portal footer makes me think of a few things.
1.  Should we text transform this so that it's all uppercase (like all other items in the portal footer)?
2. The meaning might be unclear for the user.  Could you show a tooltip on hover?  Can have the same popover info as synapse.org:

> This mode gives you early access to features that are still in development. Please note that we do not guarantee an absence of errors, and that the data created using these features may be lost during product upgrade.

emmanmills: Upgrading SRC to release 1.16.0 will pick up the tooltip and uppercase changes in this [PR](https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1037).

jay-hodgson (approved):

PR #615, opened 2021-05-21
SWC-5612 - Use SynapseContextProvider so SRC components receive token via context

nickgros:
Depends on https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1047

Will need an update to SRC to get those changes

### DIFF
--- a/src/AppInitializer.tsx
+++ b/src/AppInitializer.tsx
@@ -6,6 +6,7 @@ import { withCookies, ReactCookieProps } from 'react-cookie'
 import { DOWNLOAD_FILES_MENU_TEXT } from 'synapse-react-client/dist/containers/table/SynapseTableConstants'
 import { UserProfile } from 'synapse-react-client/dist/utils/synapseTypes'
 import SynapseRedirectDialog from 'portal-components/SynapseRedirectDialog'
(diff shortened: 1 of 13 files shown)


src/AppInitializer.tsx
```diff
@@ -145,7 +143,14 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
       return <></>
     }
     return (
-      <TokenContext.Provider value={this.state.token}>
+      <SynapseWrapper
```
nickgros, line 146: This is where the magic happens

src/RouteResolver.tsx
```diff
@@ -73,9 +76,14 @@ export const generateSynapseObject = (
...
-    <TokenContext.Consumer key={key}>
-      {(value: string) => {
-        const propsWithSearchAndToken = { ...props, searchParams, token: value }
+    <SynapseContextConsumer key={key}>
+      {(ctx?: SynapseContextType) => {
+        const propsWithSearchAndToken = {
+          ...props,
+          searchParams,
+          token: ctx?.accessToken,
+          accessToken: ctx?.accessToken,
```
nickgros, line 85: Some components (in both SRC and portals) still take `token` or `accessToken` as a prop, so we cannot remove this yet.

jay-hodgson (approved):

PR #630, opened 2021-06-25
PORTALS-1948:  Add new Download Cart page to all portals

jay-hodgson:
Add new header nav bar item ShowDownloadV2, which only shows up in alpha mode.  Adjust the user dropdown menu Downloads item to go to this new route (if in alpha mode).


jay-hodgson:
Screen Shot 2021-06-25 at 9 56 00 AM
Screen Shot 2021-06-25 at 9 56 09 AM

PR #636, opened 2021-07-01
Mobile usability issue reported by Google Search Console (about nf, but all sites probably suffer from it)

jay-hodgson:
Let's hide the Experimental Mode component in the footer if on mobile (it's extending the viewport)


jay-hodgson:
(I'll follow up with any code review changes, but I think this might be blocking Abby's build from succeeding, and they want to re-release that portal by Friday.  So merging...)

PR #661, opened 2021-08-18
Upgrade material-ui and SRC

nickgros:
Fixes env error where old versions of material core conflict with material lab used by SRC


jay-hodgson (approved):

jay-hodgson:
Note that this will pull in fixes for PORTALS-1962 and PORTALS-1973

PR #670, opened 2021-09-09
PORTALS-2001

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/PORTALS-2001)

- Rename AD portal's Explore Experimental Tools to Experimental Models
- Add RedirectWithQuery as a portal component, which allows us to maintain a route with a redirect
- Changed RouteResolver
	- Use a shared route config. Essentially, I changed the route to '/DownloadCart' from custom code to shared configuration.
	- Return 'typical' pathname in addition to route when parsing pathname
	- Handle failure cases
	  - If we can't find a base route, redirect to the homepage (leveraging the shared route configuration)
	  - If we can't find a nested page, fall back to the parent
	  - If we route to a nested page without a SynapseConfig (i.e., we have nothing to render on that page), route to a descendant that has a SynapseConfig
- Small fix to hide the renamed route from the navbar


src/RouteResolver.tsx
```diff
@@ -51,22 +39,48 @@ export const getRouteFromParams = (pathname: string) => {
...
+  // This isn't typical, but just so we can load something, we (recursively) get the first child until we find a route with a SynapseConfigArray
+  while (isEmpty(route.synapseConfigArray) && route.isNested) {
+    route = route.routes[0]
+    routePathName += '/' + route.to
+  }
+
+  return [route, routePathName]
 }
 
-export const generateSynapseObjectHelper = (synapseConfig: SynapseConfig) => {
```
nickgros, line 69: Renamed and now called through JSX since this is a function that unconditionally renders a React component

src/RouteResolver.tsx
```diff
@@ -86,44 +100,47 @@ export const generateSynapseObjectHelper = (synapseConfig: SynapseConfig) => {
   }
 }
 
-export const generateSynapseObject = (
```
nickgros, line 89: Renamed and now called through JSX since this is a function that unconditionally renders a React component

src/RouteResolver.tsx
```diff
@@ -51,22 +39,48 @@ export const getRouteFromParams = (pathname: string) => {
       split.push('')
     }
   }
-  let route = routesConfig.find((el) => split[0] === el.to)!
+  let route = ([...sharedRouteConfig, ...routesConfig] as GenericRoute[]).find(
+    (el) => split[0] === el.to,
+  )!
```
nickgros, line 44: Pull in the shared configuration

src/RouteResolver.tsx
```diff
@@ -51,22 +39,48 @@ export const getRouteFromParams = (pathname: string) => {
...
+
+  if (route == null) {
+    console.warn(`Error: url at ${pathname} has no route mapping`)
+    return [homeRoute, '/']
+  }
```
nickgros, line 49: If there's no base route, reroute to home

src/RouteResolver.tsx
```diff
@@ -51,22 +39,48 @@ export const getRouteFromParams = (pathname: string) => {
...
+      const nextRoute = route.routes.find((el) => el.to!.includes(split[i]))
+      if (nextRoute) {
+        route = nextRoute
+        routePathName += '/' + route.to
+      } else {
+        // If we can't find a matching nested route, return the last route that was found
+        break
+      }
     }
   }
```
nickgros, line 65: Gross loop that involves a `break` statement. Might be easier to understand if you look at the test cases

src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -333,6 +360,23 @@ const routes: GenericRoute[] = [
...
+            name: 'RedirectWithQuery',
+            props: {
+              exact: false,
+              strict: false,
+              from: 'Experimental Tools',
+              to: 'Experimental Models',
+            },
+          },
+        ],
+      },
```
nickgros, line 376: This lets us do the redirect!

src/RouteResolver.tsx
jay-hodgson: Should probably output a warning to the console about the sub-route that can't be found.

src/RouteResolver.tsx
```diff
@@ -51,22 +39,48 @@ export const getRouteFromParams = (pathname: string) => {
...
-  return route
+
+  // If there's no SynapseConfigArray, then the route we settled on wasn't meant to be a standalone page, and we have nothing to render.
+  // This isn't typical, but just so we can load something, we (recursively) get the first child until we find a route with a SynapseConfigArray
+  while (isEmpty(route.synapseConfigArray) && route.isNested) {
```
jay-hodgson, line 69: This part didn't make sense until I looked at the use case described by the test (/Explore/Not-A-Valid-Route -> /Explore/Data).  Nice catch.

jay-hodgson (approved):

PR #674, opened 2021-09-22
PORTALS-2015

nickgros:
https://user-images.githubusercontent.com/17580037/134384545-1ee79f8a-aef2-4d11-a388-fbc9e409cf3a.mov


src/utils.ts
```diff
@@ -1,6 +1,31 @@
...
+      }
+    }
+    window.addEventListener('resize', listener)
+    return () => {
+      window.removeEventListener('resize', listener)
+    }
+  })
+
+  return showDesktop
+}
```
nickgros, line 31: Pulled from SRC

jay-hodgson: Should this be exported from SRC instead?  Can usedBreakpoint be a const?

jay-hodgson (approved):

src/utils.ts
nickgros: In general, I think I'd rather pull in a library like [this](https://github.com/streamich/react-use) than export any non-Synapse-related utility hooks from SRC and maintain them ourselves--I'll go ahead and make that change

PR #677, opened 2021-09-27
Portals-2017

allaway:
Adding additional DOI to markdown configuration as per comment in https://sagebionetworks.jira.com/browse/PORTALS-2017


jay-hodgson (changes requested):
Very close!  The labelLinkConfig is an array of config objects, so need to slightly tweak this.  I'll create a PR based on your proposed changes.  Thank you!

jay-hodgson:
https://github.com/Sage-Bionetworks/portals/pull/678

allaway:
Thanks Jay!

PR #679, opened 2021-09-29
Style tweaks

nickgros:
Make the tab scroll button icons slightly bigger and apply the portals accent color

image

### DIFF
--- a/src/_Core.scss
+++ b/src/_Core.scss
@@ -203,6 +203,10 @@ body {
   &.explore-nav {
     margin-left: -24px;
     height: unset;
+    .MuiTabScrollButton-root .MuiSvgIcon-root {
+      color: Portal.$secondary-action-color;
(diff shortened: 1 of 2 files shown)


jay-hodgson (approved):

src/_Core.scss
```diff
@@ -203,6 +203,10 @@ body {
   &.explore-nav {
     margin-left: -24px;
     height: unset;
+    .MuiTabScrollButton-root .MuiSvgIcon-root {
```
jay-hodgson, line 206: You might be able to add a custom css class (in case MUI decides to rename their classes in the future).
Like this:  https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1246/files#diff-71e15e034ccd52858f944a7e94be6963cbabff2612660573f8ceaab9609a85b0R6

PR #682, opened 2021-10-01
update schema for tool cards

vpchung:
The tools schema for [Cancer Complexity KP](https://cancercomplexity.synapse.org/) has been updated to use new columns and annotations -- new schema [here](https://www.synapse.org/#!Synapse:syn26127427/tables/).

Note:
Other than running the tool locally on my machine, I did not run any other tests.  Happy to report that nothing broke when the updates were implemented though :)


src/configurations/cancercomplexity/resources.ts
```diff
@@ -4,4 +4,4 @@ export const filesSql =
 export const grantsSql = `SELECT * FROM syn21918972`
 export const projectsSql = `SELECT * FROM syn21868602`
 export const publicationSql = 'SELECT * FROM syn21868591'
-export const toolsSql = 'SELECT * FROM syn21930566'
+export const toolsSql = 'SELECT * FROM syn26127427'
```
nickgros, line 7: Need to make this ['Open Data'](https://sagebionetworks.jira.com/wiki/spaces/PLFM/pages/792231937/Configuring+Tables+to+Allow+Anonymous+View+Access) before we can merge this change to production

vpchung: Thanks!  Just [filed a JIRA](https://sagebionetworks.jira.com/browse/PLFM-6949) for Xa to review and approve.

src/configurations/cancercomplexity/synapseConfigs/tools.ts
```diff
@@ -15,11 +15,13 @@ export const toolsConfiguration: CardConfiguration = {
...
+      'inputFormat',
+      'outputData',
+      'outputFormat',
+      'language',
       'toolType',
-      'grantName',
+      'grantNumber',
       'consortium',
     ],
     link: 'homepageUrl',
```
nickgros, line 27: Seems like the new column with this data is now named `homepage`

vpchung: Nice catch!  Pushing a fix.

vpchung:
@nickgros I think I found another breaking point.  When I go to a Grant card, none of the related tools are showing:

Screen Shot 2021-10-04 at 9 34 25 AM

which isn't encountered in prod:

Screen Shot 2021-10-04 at 9 34 37 AM

I'll keep sleuthing.

nickgros:
> @nickgros I think I found another breaking point. When I go to a Grant card, none of the related tools are showing:

@vpchung This is definitely caused by a mismatch in the [routesConfig](https://github.com/Sage-Bionetworks/portals/blob/fa6da82e1ae2d7e05fe8c210b6f6d41f29832370/src/configurations/cancercomplexity/routesConfig.ts#L226). For just this issue, I think it would suffice to just change `grantName` to `grantNumber` in the "Related Tools" object, but it would probably be a good idea for us to look at the other Details Page configs too to make sure they are still accurate w.r.t. the schema you're using.

vpchung:
@nickgros ooooh that's where it is.  I kept looking in `grants.ts` but nothing stood out, thanks!

I'm worried that changing to `grantNumber` will break more things downstream, so maybe it'll be better to add `grantName` back into the schema?  I'll confirm with @jaeddy.

vpchung:
I have added `grantName` back into the new table, and have updated the config files accordingly.  Now things are working as expected :)

We also only want to display certain tools from the table, hence why the query has been updated as well.

nickgros (approved):
Looks good to me! Thanks @vpchung!

PR #687, opened 2021-10-11
PORTALS-2020: set the ShowDownloadV2 class name to look like a nav bar item

jay-hodgson:


jay-hodgson:
Screen Shot 2021-10-11 at 3 40 28 PM

Mobile improved, but still not great:
Screen Shot 2021-10-11 at 3 40 45 PM

PR #690, opened 2021-10-15
Expose id column

kelshmo:
Given the change to the id column arriving with stack 379: [SWC-5806](https://sagebionetworks.jira.com/browse/SWC-5806), this PR: 
- fixes the facet alias, id is now actually the id and the name column can be faceted as "File Name"
- updated the metadataSql to AD standards by using the resourceType key
- increments the data view to test these changes

### DIFF
--- a/src/configurations/psychencode/resources.ts
+++ b/src/configurations/psychencode/resources.ts
@@ -1,7 +1,7 @@
 // Portal owners can change the versions of the resources by modifying the
 // version of the entity used in the sql below
 
-const dataSynId = 'syn20821313.13'
+const dataSynId = 'syn20821313.14'
 export const studiesSql = 'SELECT * FROM syn21783965.9'
 export const dataSql = `SELECT * FROM ${dataSynId}`
 export const metadataSql = `SELECT id, metadataType, dataType, assay FROM ${dataSynId} WHERE "resourceType" = \'metadata\'`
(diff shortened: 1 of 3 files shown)


jay-hodgson:
@kelshmo : [SWC-5806](https://sagebionetworks.jira.com/browse/SWC-5806) is specific to Synapse.org (and only to Entity Views), and does not apply to portals.  Duminda was very clear that they wanted to leave Portals completely unchanged.  
And because I took so long to review this PR, it now has merge conflicts in src/configurations/psychencode/synapseConfigs/studies.ts (which you may want to revert your changes to this file anyway, with the info above).

kelshmo:
@jay-hodgson I'm confused since the available data table is driven by a view. So, when I create a new view version and make it live on the portal site, the change will be there. (It already is there if you look at the most recent version of syn20821313).

jay-hodgson:
Talked with @kelshmo, and confirmed that this was not the intended output in the portal (result of this PR):
Screen Shot 2021-11-02 at 8 58 41 AM

It's important to note that the data is unchanged, only the rendering was changed on Synapse.org.  
The View ID column still contains Synapse IDs (though rendered as Entity Names on the portals, and Synapse IDs on Synapse.org), and will continue to be exported as Synapse IDs.

PR #700, opened 2021-11-17
PORTALS-2028 (part 2): add a redirect for MODEL-ADstrains, and supply…

jay-hodgson:
… the search parameters in the config

### DIFF
--- a/src/configurations/adknowledgeportal/routesConfig.ts
+++ b/src/configurations/adknowledgeportal/routesConfig.ts
@@ -31,6 +31,7 @@ import {
   peopleSql,
   programsSql,
   experimentalModelsSql,
+  modelADStrainsSelectedFacet,
 } from './resources'
 
 const routes: GenericRoute[] = [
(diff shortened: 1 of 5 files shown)


src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -215,7 +216,21 @@ const routes: GenericRoute[] = [
...
+    // PORTALS-2028 (part 2): redirect /MODEL-ADstrains to /Explore/Experimental%20Models with query request
+    isNested: false,
+    to: 'MODEL-ADstrains',
+    hideRouteFromNavbar: true,
+    synapseConfigArray: [
+      {
+        name: 'Redirect',
+        props: {
+          from: 'MODEL-ADstrains',
+          to: { pathname: '/Explore/Experimental Models', search: `QueryWrapper0={"sql":"${encodeURI(experimentalModelsSql)}","limit":25,"offset":0,"selectedFacets":[{"concreteType":"org.sagebionetworks.repo.model.table.FacetColumnValuesRequest","columnName":"program","facetValues":["MODEL-AD"]}]}` },
```
nickgros, line 229: Would it be a good idea to co-locate the facet query with the `experimentalModelsSql` constant so that if the table updates, it's more likely that we remember to update both?

nickgros (approved):

PR #712, opened 2022-01-03
PORTALS-2063: if showMenu is true (is the default) then also show the…

jay-hodgson:
… left-hand nav menu on tabbed detail pages

### DIFF
--- a/src/portal-components/DetailsPage.tsx
+++ b/src/portal-components/DetailsPage.tsx
@@ -198,11 +198,12 @@ const handleMenuClick = (
     console.error('Could not scroll to element with index ', index)
   }
 }
-const renderMenu = (
-  ref: React.RefObject<HTMLDivElement>,
-  synapseConfigArray?: RowSynapseConfig[],
+
+const SideNavMenu: React.FC<{
+  clickRef: React.RefObject<HTMLDivElement>
+  synapseConfigArray?: RowSynapseConfig[]
   queryResultBundle?: QueryResultBundle
-) => {
+}> = ({ clickRef, synapseConfigArray, queryResultBundle }) => {
   const mapColumnHeaderToRowIndex: Dictionary<number> = {}
   let row: string[] = []
   if (queryResultBundle) {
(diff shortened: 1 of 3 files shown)


src/portal-components/DetailsPage.tsx
```diff
@@ -56,7 +56,6 @@ export default class DetailsPage extends React.Component<
   DetailsPageProps,
   State
 > {
-  public ref: React.RefObject<HTMLDivElement>
```
jay-hodgson, line 59: Moved ref to DetailsPageSynapseConfigArray

src/portal-components/DetailsPage.tsx
```diff
@@ -110,26 +108,6 @@ export default class DetailsPage extends React.Component<
     }
   }
 
-  handleMenuClick = (index: number) => {
```
jay-hodgson, line 113: Moved

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
       return (
         <div className="DetailsPage tab-layout">
-          <div className="component-container" ref={this.ref}>
+          <div className="component-container">
             {
               <DetailsPageTabs
                 tabConfigs={tabLayout}
                 loading={isLoading}
                 queryResultBundle={this.state.queryResultBundle}
+                showMenu={showMenu}
```
jay-hodgson, line 157: Pass through

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
               ></DetailsPageTabs>
             }
           </div>
         </div>
       )
     } else {
-      const synapseConfigContent = (
```
jay-hodgson, line 185: DetailsPageSynapseConfigArray now handles menu (if showMenu is true), so moved this logic

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
-          </div>
-        )
-      } else {
-        return synapseConfigContent
-      }
     }
   } // end render()
+}
 
-  renderMenu = () => {
```
jay-hodgson, line 211: Moved

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
+      (el, index) => {
+        mapColumnHeaderToRowIndex[el.name] = index
+      },
+    )
+    row = queryResultBundle.queryResult.queryResults.rows[0].values
+  }
+  return (
+    synapseConfigArray &&
+    synapseConfigArray.map((el: RowSynapseConfig, index) => {
+      if (!el.title) {
```
jay-hodgson, line 219: This is a conditional check that I added.  If the config has no title, then do not include it in the left-hand side menu.

jay-hodgson:
When showMenu is true (or not specified):
Screen Shot 2022-01-03 at 3 26 09 PM

No change to detail pages that do not have tabs:
Screen Shot 2022-01-03 at 3 30 42 PM

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
-            {el.title}
-          </button>
+            synId={el.props.synId}
+          />
         )
-      })
-    )
-  }
+      }
+      if (!el.title) {
```
jay-hodgson, line 240: This is a conditional check that I added. If the config has no title, then do not include it in the left-hand side menu.

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
+    const offsetPosition = elementPosition - offset
+    window.scrollTo({
+      top: offsetPosition,
+      behavior: 'smooth',
+    })
+  } else {
+    console.error('Could not scroll to element with index ', index)
+  }
+}
+const renderMenu = (
```
jay-hodgson, line 201: Terrible diff, renderMenu() was moved, with the addition of one conditional check (noted below)

src/portal-components/_DetailsPage.scss
```diff
@@ -81,9 +81,7 @@ $svg-icon-height: 30px;
   }
 
   &.tab-layout {
-    .component-container {
```
jay-hodgson, line 84: This did not play well with the sticky menu (component content was moved up, right next to the tab).  
component-container sets the layout, but tab-layout is responsible for moving the tabs up (into the header card area).

src/portal-components/DetailsPage.tsx
```diff
@@ -170,94 +148,110 @@ export default class DetailsPage extends React.Component<
...
+  ref: React.RefObject<HTMLDivElement>,
+  synapseConfigArray?: RowSynapseConfig[],
+  queryResultBundle?: QueryResultBundle
+) => {
```
nickgros, line 205: I don't think there are any issues here since this component has no hooks, but would prefer treating this as a component since it unconditionally returns a React element, then render it with JSX instead of calling it.

nickgros:
Not sure how I'm missing it but I don't see where the default value for `showMenu` is set (and I'm sure that it is, or else there would be a TS compile error)

nickgros (approved):
Looks great, much cleaner!

jay-hodgson:
> Not sure how I'm missing it but I don't see where the default value for `showMenu` is set (and I'm sure that it is, or else there would be a TS compile error)
@nickgros : They seem to set the default when pulling showMenu out of the props in DetailsPage.render():
const { showMenu = true, tabLayout, synapseConfigArray } = this.props

nickgros (approved):

PR #718, opened 2022-01-13
PORTALS-2088: use an override link column (biobankURL)

jay-hodgson:


jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1350

jay-hodgson:
Must update SRC to a new version as well.

jay-hodgson:
Note that this PR will have no effect until the [underlying tools table](https://www.synapse.org/#!Synapse:syn26438037) has the override link URL column (currently looking for "biobankURL").

PR #720, opened 2022-01-20
Modify publicationsV2Sql...

allaway:
...to use Usage_Publications joined table. 

Poking around, it seems like this query is only used for the Publications section on the Tool Detail pages, so we should be able to just do a straight swap, but let me know if I've missed something! :)


jay-hodgson (approved):
Checked out this branch and ran the nf portal locally, verified that the publication v2 sql update does not break the Tools detail page (tested with a tool that has associated publications).

PR #721, opened 2022-01-26
PORTALS-2057: avoid pushing a path into the history that also has a s…

jay-hodgson:
…earch part

Also fix the About page in NF (looks like it changed entity owner?)


nickgros (commented):
Great sleuthing

PR #723, opened 2022-02-03
PORTALS-2096

jay-hodgson:
Rework Synapse redirect dialog to use a react bootstrap Modal instead of a MUI Dialog.
Implement way to send the cancel message to AppInitializer in order to clear the redirect URL.


jay-hodgson:
On cancel, I also reset the countdown.
Screen Shot 2022-02-03 at 3 57 32 PM

PR #730, opened 2022-02-25
PORTALS-2101

yjyoo773:


src/Navbar.tsx
```diff
@@ -9,6 +9,9 @@ import NavLink from 'portal-components/NavLink'
 import NavUserLink from './portal-components/NavUserLink'
 import { GenericRoute } from 'types/portal-config'
 import Button from 'react-bootstrap/esm/Button'
+import IconButton from '@material-ui/core/IconButton'
```
nickgros, line 12: @yjyoo773 from the issue comments, I thought we decided to convert this to react-bootstrap. Is there a reason you decided to continue using Material-UI?

src/AppInitializer.tsx
```diff
@@ -114,7 +114,8 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
     })
   }
 
-  handleCloseLoginDialog = () => {
+  handleCloseLoginDialog = (event, reason) => {
+    if(reason && reason === "backdropClick") return
```
nickgros, line 118: Is this still necessary now that we have `backdrop='static'`?

yjyoo773: No it isn't. I'll remove this

src/Navbar.tsx
```diff
@@ -192,15 +192,15 @@ class Navbar extends React.Component<any, State> {
...
-                  >
-                    <SynapseComponents.Login
-                      sessionCallback={() => getSession()}
-                    />
-                  </Dialog>
+                  <Modal 
+                    animation={false}
+                    backdrop={'static'}
+                    onHide={handleCloseLoginDialog} 
+                    style={{'marginTop':'5%'}}
```
nickgros, line 199: Would prefer this to be a number (px) in case the width of this component changes in the future.

nickgros (changes requested):
Small questions/changes requested but overall looks good!

nickgros (approved):

PR #739, opened 2022-03-21
Tool cards and details update

vpchung:
The CSBC/PS-ON team would like to display more information for the Tool components, now that we have transitioned over to a new schema.

Notable changes:
* New labels found on a Tool card
* Clicking on a Tool card now redirects to a Tool Details page (rather than redirecting to the tool itself)

Some whitespace were also removed (via autoformatter).

[Related JIRA ticket](https://sagebionetworks.jira.com/browse/CPO-10)


nickgros (approved):
Looks great to me!

PR #745, opened 2022-04-19
Add configurations for new portal table in CCKP

vpchung:
The CCKP is adding a new table to the portal ([Portal - People Merged](https://www.synapse.org/#!Synapse:syn28073190/tables/)) which will list related persons from the CSBC / PS-ON projects.  The table is currently being reviewed by Governance for enabling anonymous download.

### FIXME ?
I have began making some adjustments to the configuration files so that the new table will be displayed on the portal.  One thing I am not able to figure out is the SQL logic in `routesConfig.ts` (lines 277-318) and how to get the card to properly map to related publications, datasets, and tools.  I believe the issue is because the keys (datasetId, publicationId, and toolId) are an EntityIdList in the People table, not String or EntityId?  For now, I have created temporary columns in the table so that the mapping will work on the portal.


vpchung:
@nickgros thank you!  And no worries - I figured that was the case. I'll let the data team know that using a list is not ideal at this time, and we will try to think of other ways of displaying the information.

Thanks again!

nickgros:
@vpchung one option could be to normalize into multiple tables and use one or more MaterializedViews to construct the appropriate results. Even though MaterializedViews are only in experimental mode, if the data in it appears properly in Synapse then we can reference it from a Portal

src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -275,6 +356,18 @@ const routes: GenericRoute[] = [
                   sql: publicationSql,
                   sqlOperator: 'LIKE',
                   synapseConfigArray: [
+                    {
+                      name: 'CardContainerLogic',
+                      columnName: 'publicationId',
+                      title: 'Related People',
+                      tableSqlKeys: ['publicationId'],
+                      props: {
+                        sqlOperator: 'LIKE',
```
jay-hodgson, line 365: Does sqlOperator '=' not work here?

src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -430,6 +523,18 @@ const routes: GenericRoute[] = [
                         facetAliases,
                       },
                     },
+                    {
+                      name: 'CardContainerLogic',
+                      columnName: 'toolId',
+                      title: 'Related People',
+                      tableSqlKeys: ['toolId'],
+                      props: {
+                        sqlOperator: 'LIKE',
```
jay-hodgson, line 532: Or here?

src/configurations/cancercomplexity/routesConfig.ts
vpchung: Oh yes, you are right - using `=` works here.

jay-hodgson (approved):

PR #753, opened 2022-05-03
Bump react scripts to v5

nickgros:
* Bump react-scripts
* Add craco, browser polyfills

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,7 @@
     "react-spinners": "^0.11.0",
     "react-use": "^17.3.1",
     "sass": "^1.30.0",
-    "synapse-react-client": "3.0.8",
-    "typescript": "4.3.5"
+    "synapse-react-client": "3.0.8"
   },
   "jest": {
     "moduleNameMapper": {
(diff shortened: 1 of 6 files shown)


src/tests/RouterResolver.test.tsx
```diff
@@ -106,7 +106,6 @@ describe('RouteResolver works', () => {
             ],
           },
         ],
-        colors: ['red'],
```
nickgros, line 109: I think this is a prop that has since been removed

.gitignore
```diff
@@ -5,4 +5,14 @@ src/config/
...
+.idea/
+
+# https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
+# We currently aren't using Zero-Installs
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
```
nickgros, line 17: All of the files in .yarn/sdks is auto-generated, but we're instructed to commit those files to SCM

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "my-app",
+  "name": "portals",
```
nickgros, line 2: Makes messages emitted by Yarn easier to interpret

src/configurations/adknowledgeportal/style/_style_overrides.scss
```diff
@@ -1,4 +1,4 @@
-@use 'variables' as Portal;
+@use '../../style/variables' as Portal;
```
nickgros, line 1: Since we're not using sass load paths anymore, our file references need to be updated. This is easier to reason with IMO.

PR #760, opened 2022-05-11
Bump react-scripts to v5

nickgros:
* Bump react-scripts
* Add craco, browser polyfills

While craco isn't currently actively maintained, I think it's still worth using it to upgrade to react-scripts v5. We may be able to just use plain react-scripts in the future, depending on whether https://github.com/facebook/create-react-app/pull/11764 (or similar PR) is merged.

### DIFF
--- a/src/Navbar.tsx
+++ b/src/Navbar.tsx
@@ -105,7 +105,12 @@ class Navbar extends React.Component<any, State> {
     } = this.props as SignInProps
     const { name, icon, hideLogin = false } = logoHeaderConfig
     const imageElement = icon ? (
-      <img id="header-logo-image" alt="navigation logo" className="nav-logo" src={icon} />
+      <img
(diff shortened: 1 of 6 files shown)


src/Navbar.tsx
```diff
@@ -192,15 +197,20 @@ class Navbar extends React.Component<any, State> {
...
-                  <Modal 
+                  <Modal
                     animation={false}
-                    backdrop='static'
+                    backdrop="static"
                     keyboard={false}
-                    onHide={handleCloseLoginDialog} 
+                    onHide={() => {
+                      handleCloseLoginDialog()
+                    }}
```
nickgros, line 206: Fixes a linting error where the type of handleCloseLoginDialog didn't match the permitted type of onHide

jay-hodgson (approved):

PR #761, opened 2022-05-16
PORTALS-1883: pubMedId column is now a String column

jay-hodgson:
(not String_list), so fix the sql operator


jay-hodgson:
The example page is now error free:
Screen Shot 2022-05-16 at 1 51 14 PM

vpchung (approved):
Looks great!

PR #773, opened 2022-06-03
SWC-6147: Add example (unlinked) Dataset explore and detail page to N…

jay-hodgson:
…F.  Currently points to a test Dataset View (to drive the Explore module) which contains a test Dataset

### DIFF
--- a/src/configurations/nf/synapseConfigs/datasetsV2.ts
+++ b/src/configurations/nf/synapseConfigs/datasetsV2.ts
@@ -1,0 +1,104 @@
+import { GenericCardSchema } from 'synapse-react-client/dist/containers/GenericCard'
+
+import { SynapseConstants } from 'synapse-react-client'
+import { SynapseConfig } from 'types/portal-config'
+import { facetAliases } from './commonProps'
+import { CardConfiguration } from 'synapse-react-client/dist/containers/CardContainerLogic'
+import {
+  filesSql,
+} from '../resources'
+import { DetailsPageProps } from 'types/portal-util-types'
+
+const datasetsSql = 'SELECT * FROM syn31543490'
+export const datasetSchema: GenericCardSchema = {
+  type: SynapseConstants.DATASET,
+  title: 'name',
+  subTitle: 'institutionOrPI', // custom annotation on datasets
+  description: 'descriptionAnnotation', // until entity.description is supported, this also needs to be a custom annotation
+  secondaryLabels: [
+    'assay', 
+    'currentVersion',
+    'createdBy',
+    'id'
+  ],
+}
+
+export const datasetCardConfiguration: CardConfiguration = {
+  type: SynapseConstants.GENERIC_CARD,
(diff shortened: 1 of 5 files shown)


jay-hodgson:
Still need to fix the Dataset icon, but I think that's in SRC.

Unlinked Explore module:
Screen Shot 2022-06-03 at 10 45 39 AM

Unlinked Details page:
Screen Shot 2022-06-03 at 10 50 27 AM

src/portal-components/DetailsPage.tsx
```diff
@@ -382,6 +384,10 @@ const SplitStringToComponent: React.FC<{
   const injectedProps = injectPropsIntoConfig(value, el, {
     ...deepCloneOfProps,
   })
+  if (overrideSqlSourceTable) {
```
jay-hodgson, line 387: Note the new boolean parameter overrideSqlSourceTable in the RowToPropTransform.  (I made this sql spacing a little strange in case we need to track it down in the future!)

src/portal-components/injectPropsIntoConfig.ts
```diff
@@ -17,6 +17,7 @@ const injectPropsIntoConfig = (
   el: RowSynapseConfig,
   props: any,
 ): any => {
+  // If isValueSourceEntity, then
```
nickgros, line 20: Is this relevant? Not sure what this means/what the alternative is

src/configurations/nf/synapseConfigs/datasetsV2.ts
```diff
@@ -0,0 +1,104 @@
...
+        visibleColumnCount: 7,
+        tableConfiguration: {
+          showAccessColumn: true,
+          showDownloadColumn: true,
+        },
+        shouldDeepLink: false,
+        facetAliases,
+      },
+      // tableSqlKeys: ['id'],  // Do not modify the sql where condition based on search params
+      overrideSqlSourceTable: true, // Instead, modify the sql (SELECT * FROM <search_param_value>)
```
nickgros, line 79: Nice, it only applies to this QueryWrapperPlotNav!!

nickgros:
Would it be a lot of effort to add a test or two in DetailsPage.test.tsx? Maybe one option would be to create a stub component that renders the value of the SQL prop and test the rendered value with/without the `overrideSqlSourceTable` prop.

nickgros (approved):
LGTM, but see comments

src/portal-components/injectPropsIntoConfig.ts
jay-hodgson: Removed

src/configurations/nf/synapseConfigs/datasetsV2.ts
jay-hodgson: Yes!

jay-hodgson:
Working on a followup PR to add a test to DetailsPage.test.tsx to verify the effects of the new param.

PR #778, opened 2022-06-07
react-router v6 pre-work

nickgros:
See https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1562

Upgrading to v6 will require much more work for portals, seems like migrating could save us a lot of code though, e.g. useRoutes looks a lot like our RouteResolver and configurations -- https://reactrouter.com/docs/en/v6/upgrading/v5#use-useroutes-instead-of-react-router-config


src/AppInitializer.tsx
```diff
@@ -247,4 +247,4 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
   }
 }
 
-export default withRouter(withCookies(AppInitializer))
+export default withLocation(withCookies(AppInitializer))
```
nickgros, line 250: withRouter is removed in react-router v6 and AppInitializer is a class component, so it can't use `useLocation`. Created a small wrapper `withLocation` that passes the location as a prop using the `useLocation` hook, which unblocks upgrading to v6.

src/CookiesNotification.tsx
```diff
@@ -45,22 +44,22 @@ class CookiesNotification extends React.Component<
...
         }}
         isGlobal={true}
-        />
+      />
     )
   }
 }
 
-export default withRouter(withCookies(CookiesNotification))
+export default withCookies(CookiesNotification)
```
nickgros, line 65: Component didn't use the router

src/Home.tsx
```diff
@@ -12,4 +11,4 @@ const Home: React.SFC = () => {
   )
 }
 
-export default withRouter(Home)
+export default Home
```
nickgros, line 14: Didn't use router data

src/Home.tsx
```diff
@@ -1,9 +1,8 @@
 import * as React from 'react'
 import { Header } from './Header'
-import { withRouter } from 'react-router-dom'
 import RouteResolver from './RouteResolver'
 
-const Home: React.SFC = () => {
+const Home = () => {
```
nickgros, line 5: Deprecated type

src/RouteResolver.tsx
```diff
@@ -197,4 +197,4 @@ const RouteResolver: React.FunctionComponent<RouteComponentProps> = () => {
   )
 }
 
-export default withRouter(RouteResolver)
```
nickgros, line 200: RouteResolver was already upgraded to use hooks in the past. withRouter is unnecessary

src/portal-components/RouteControlWrapper.tsx
```diff
@@ -88,4 +88,4 @@ const RouteControlWrapper: React.FunctionComponent<Props> = ({
   )
 }
 
-export default withRouter(RouteControlWrapper)
+export default RouteControlWrapper
```
nickgros, line 91: useLocation and useHistory replace withRouter

src/withLocation.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export type WithLocationPropType = {
+  location: Location
+}
+
+export default function withLocation(
+  Component: React.ComponentType<WithLocationPropType>,
+) {
+  return (props) => <Component {...props} location={useLocation()} />
+}
```
nickgros, line 12: Just passes location data from the hook. Allows us to replace `withRouter` for `AppInitializer`

jay-hodgson (approved):

PR #782, opened 2022-06-13
PORTALS-2178

jay-hodgson:


jay-hodgson:
In testing, I changed resources.ts toolsSql to select from syn31927658 (Robert's test table).
It resulted in the following output (note that the text in the cell is run through the markdown processor, since I saw the possibility for links in the design):
Screen Shot 2022-06-13 at 2 51 50 PM

PR #783, opened 2022-06-13
PORTALS-2177

jay-hodgson:


jay-hodgson:
In testing, I changed resources.ts toolsSql to select from syn31927658 (Robert's test table).
It resulted in the following output:
Screen Shot 2022-06-13 at 3 06 03 PM

jay-hodgson:
Merged upstream master and tested one tools detail page.
Screen Shot 2022-06-15 at 5 26 48 PM

PR #786, opened 2022-06-15
SWC-6127: with a View Sharing Settings button that takes you to the c…

jay-hodgson:
…orresponding entity page (on the Synapse production website)


jay-hodgson:
The download permission required action should be an edge case in a data dissemination portal (since the data is usually governed via Access Requirements), but the button should do something.
Screen Shot 2022-06-15 at 9 06 53 AM

PR #792, opened 2022-07-01
add mapping from People -> Grant

vpchung:
Requested by @aclayton555 :

> also wondering is if the Grant Name on the People Details pages can also link to the respective Grant Details page - similar to how Publications, etc Details pages do?

QC'd on my end and everything looks okay so far!

**Screenshots**
person_with_grant

person_with_mult_grants

### DIFF
--- a/src/configurations/cancercomplexity/routesConfig.ts
+++ b/src/configurations/cancercomplexity/routesConfig.ts
@@ -205,7 +205,7 @@ const routes: GenericRoute[] = [
                       title: 'Related People',
                       tableSqlKeys: ['grantNumber'],
                       props: {
-                        sqlOperator: 'LIKE',
+                        sqlOperator: 'HAS',
                         sql: peopleSql,
                         ...peopleCardConfiguration,
                         facetAliases,
--- a/src/configurations/cancercomplexity/routesConfig.ts
+++ b/src/configurations/cancercomplexity/routesConfig.ts
@@ -280,6 +280,18 @@ const routes: GenericRoute[] = [
                   synapseConfigArray: [
                     {
                       name: 'CardContainerLogic',
+                      columnName: 'grantNumber',
+                      title: 'Related Grants',
+                      tableSqlKeys: ['grantNumber'],
+                      props: {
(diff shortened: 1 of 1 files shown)


vpchung:
@aclayton555 also requested:

> wondering if, similarly, we could add “Related People” to Grant details pages - although i am not sure if it is easy for the mapping to go that way?

For this mapping, I decided to use the grant numbers as the mapping key (rather than names, since first and last names might not be listed in the same order between grants).  Which works great still!

**Screenshot**
Screen Shot 2022-07-01 at 11 53 46 AM

src/configurations/cancercomplexity/routesConfig.ts
```diff
@@ -199,6 +199,18 @@ const routes: GenericRoute[] = [
                         ...projectCardConfiguration,
                       },
                     },
+                    {
+                      name: 'CardContainerLogic',
+                      columnName: 'grantNumber',
+                      title: 'Related People',
+                      tableSqlKeys: ['grantNumber'],
+                      props: {
+                        sqlOperator: 'LIKE',
```
jay-hodgson, line 208: Since grantNumber is a StringList in the [people table](https://www.synapse.org/#!Synapse:syn28073190/tables/), does the sqlOperator 'HAS' work here?

vpchung: You're right!

I'm curious - is there a performance difference between using `LIKE` and `HAS`?

jay-hodgson: @marcomarasca can probably provide more info.  My understanding is that it should have improved database performance (HAS should be looking for the exact string match), but probably not noticeable for this amount of data.

jay-hodgson: @vpchung : If HAS works here, could you commit that change (and I will approve/merge the PR)?

vpchung: Done! Sorry for the wait.

jay-hodgson (approved):

PR #802, opened 2022-07-12
PORTALS-2228

nickgros:
* Add support for linking to sections within a DetailsPage tab
	* Details page tab selection is now reflected in the URL path	
* Refactor RouteResolver and route configurations across all portals
	*  See note below
* Refactor tests 
	* Prevent fetching data from Synapse where possible
	* Run with minimal warnings
	* Remove Enzyme

<hr/>

Before this change, we had a method [`getRouteFromParams`](https://github.com/Sage-Bionetworks/portals/blob/4fbec5d473c6d378d17cfcbc74cb3a0ed700b91e/src/RouteResolver.tsx#L19), which would selectively render content from the route tree. While we used react-router at the top level of the app, we did not use react-router at all to implement `getRouteFromParams`. Because of this, react-router could not be easily used in more deeply nested places in the app (for example, within a DetailsPage). Any additional routing would require increasingly complex custom solutions.

getRouteFromParams wasn't doing anything that couldn't be done with react-router, so this change removes it and restructures routes to be more compatible with react-router. I made major changes to the interface for route configurations to more clearly reflect react-router options, such as removing `isNested` and adding `exact`. 

<hr/>

https://user-images.githubusercontent.com/17580037/178541785-3a61dc61-204f-4093-93e2-eb045c17087a.mov


src/App.tsx
```diff
@@ -22,14 +21,7 @@ const App: React.FC = () => {
...
             <React.Suspense fallback={<div />}>
               <Switch>
-                {/* exact takes precendence over RouteResolver */}
-                <Route exact={true} path="/">
-                  <Home />
-                </Route>
-                {/* all other routes handled programatically */}
-                <Route path="/">
-                  <RouteResolver />
-                </Route>
```
nickgros, line 32: All routes now handled by RouteResolver, including Home.

src/Home.tsx
```diff
@@ -1,14 +0,0 @@
-import * as React from 'react'
-import { Header } from './Header'
-import RouteResolver from './RouteResolver'
-
-const Home = () => {
-  return (
-    <div>
-      <Header />
-      <RouteResolver />
```
nickgros, line 9: Didn't need another "RouteResolver" in the tree. Since this is just rendering a Header, let's just remove this entire component.

src/Navbar.tsx
```diff
@@ -128,9 +128,16 @@ class Navbar extends React.Component<any, State> {
...
-    })[0]
+    const homeRouteConfig: ConfigRoute | undefined = routesConfig.filter(
+      (r: GenericRoute): r is ConfigRoute => {
+        return !!(
+          r.path === '' &&
+          r.synapseConfigArray &&
+          r.synapseConfigArray.length > 0
+        )
+      },
+    )[0]
```
nickgros, line 139: Small bugfix: this renders buttons in the dropdown that link to sections of the home page. Check the length of the synapseConfigArray to see if there are actually any sections. Otherwise, the Home button will act as a dropdown with no menu items. For example, see STOP-AD portal.

src/Navbar.tsx
```diff
@@ -300,37 +307,41 @@ class Navbar extends React.Component<any, State> {
...
+                    (el.routes &&
+                      el.routes.every((route) => route.hideRouteFromNavbar))
+                  if (!el.exact && !hideChildren) {
+                    const isSelected =
+                      el.routes &&
+                      el.routes.some(
+                        (route) =>
+                          this.getLinkHref(route, topLevelTo, false) ===
+                          decodeURIComponent(window.location.pathname),
+                      )
```
nickgros, line 320: Update to use `exact` instead of `isNested`

src/Navbar.tsx
```diff
@@ -340,30 +351,32 @@ class Navbar extends React.Component<any, State> {
                             {displayName}
                           </Dropdown.Toggle>
                           <Dropdown.Menu className="portal-nav-menu">
-                            {el.routes.map((route) => {
```
nickgros, line 343: Add type checking. Other changes in this block are whitespace.

src/App.scss
```diff
@@ -16,7 +16,7 @@
 @use 'ExploreMenu';
 @use 'SynapseRedirectDialog';
 @use 'ToggleSynapseObjects';
-@use '../portal-components/DetailsPage';
+@use '../portal-components/DetailsPage/DetailsPage';
```
nickgros, line 19: Moved DetailsPage to its own folder

src/RouteResolver.tsx
```diff
@@ -1,107 +1,150 @@
...
 import Layout from './portal-components/Layout'
-import sharedRouteConfig, { homeRoute } from './shared-config/sharedRoutes'
+import sharedRouteConfig from './shared-config/sharedRoutes'
 
-/**
- * Given a pathname, return the matching route object and the route's pathname.
- * @param pathname
- * @returns
- */
-export const getRouteFromParams = (
```
nickgros, line 19: Removed; let's use react-router instead.

src/RouteResolver.tsx
```diff
@@ -1,107 +1,150 @@
...
+      ? `${docTitleConfig.name} - ${pageName}`
+      : getPageNameFromParentRoute()
+
+  // if there are children, don't update the title
+  if (!route.routes || route.routes.length === 0) {
+    const newTitle: string = getPageName()
+    if (document.title !== newTitle) {
+      document.title = newTitle
+    }
   }
```
nickgros, line 113: Defer updating the document title until we render the most specific match in the tree. If that match doesn't have an appropriate title, fall back to the parent's title.

src/RouteResolver.tsx
```diff
@@ -1,107 +1,150 @@
...
+            >
+              <RecursiveRouteRenderer
+                route={r}
+                getPageNameFromParentRoute={getPageName}
+              />
+            </Route>
+          )
+        })}
+    </>
+  )
```
nickgros, line 139: If this "node" has components, render them.

If it has children, render them recursively within a Route.

src/RouteResolver.tsx
```diff
@@ -1,107 +1,150 @@
...
-  )!
+const ROUTES: GenericRoute = {
+  ...sharedRouteConfig,
+  routes: [...(sharedRouteConfig.routes ?? []), ...routesConfig],
+}
 
-  if (route == null) {
-    console.warn(`Error: url at ${pathname} has no route mapping`)
-    return [homeRoute, '/']
+function ComponentRenderer(props: { config: SynapseConfig }) {
```
nickgros, line 17: This just renders the synapseConfigArray. Was essentially just pulled out of the previous implementation with no major changes.

src/SynapseComponent.tsx
```diff
@@ -1,19 +1,45 @@
...
 
 type SynapseComponentWithContextProps = {
   synapseConfig: SynapseConfig
   searchParams?: any
 }
 
-export const SynapseComponentWithContext: React.FC<
-  SynapseComponentWithContextProps
-> = ({ synapseConfig, searchParams }) => {
+export const SynapseComponent: React.FC<SynapseComponentWithContextProps> = ({
```
nickgros, line 39: Rename to SynapseComponent, still including context.

src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -35,8 +35,9 @@ import {
 
 const routes: GenericRoute[] = [
   {
-    to: '',
-    isNested: false,
+    path: '',
+    hideRouteFromNavbar: true,
+    exact: true,
```
nickgros, line 40: Removed `isNested`, which had custom, undocumented behavior that I'm not sure I could accurately explain. Replaced with `exact`, which comes from [react-router](https://v5.reactrouter.com/web/api/Route/exact-bool), so we get the benefits of react-router documentation.

src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -239,33 +240,35 @@ const routes: GenericRoute[] = [
     ],
   },
   {
-    to: 'Explore',
-    isNested: true,
+    path: 'Explore',
     routes: [
```
nickgros, line 244: In the type definitions, we now enforce that
* A route can only have a `synapseConfigArray` or child `routes`, but not both
* If a route has `routes`, `exact` is implicitly false, and cannot be overridden.

With how the new RouteResolver works, I can't think of a reason where you would want `exact` to be true and render child routes, because then the child routes could never render.

src/configurations/adknowledgeportal/routesConfig.ts
```diff
@@ -239,33 +240,35 @@ const routes: GenericRoute[] = [
...
-              ...RouteControlWrapperProps,
-              synapseConfig: {
-                name: 'CardContainerLogic',
+            path: '',
+            exact: true,
+            synapseConfigArray: [
+              {
+                name: 'RouteControlWrapper',
+                isOutsideContainer: true,
+                className: 'ProgramCardList',
```
nickgros, line 255: The /Explore/Programs page is a good example of the new way to render content on a page that had child routes.

Before, Programs had a `synapseConfigArray`, which rendered the RouteControlWrapper and CardContainerLogic query view. Programs also had `routes` which rendered the details page.

After these changes, we can only have `synapseConfigArray` or `routes`, but not both. So now Programs has `routes`, and the previous synapseConfigArray is included in a route with an empty string path. This new route can be shown/hidden on child pages by specifying `exact`, so this gives us a little more flexibility while having closer behavior to react-router.

src/configurations/adknowledgeportal/synapseConfigs/studies.ts
```diff
@@ -81,6 +81,7 @@ export const studiesDetailsPageProps: DetailsPageProps = {
   tabLayout: [
     {
       title: 'Study Details',
+      uriValue: 'StudyDetails',
```
nickgros, line 84: These are DetailsPage tabs, so navigating to the tab will give a URL like /Explore/Studies/DetailsPage/StudyDetails?key=value

src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -1,44 +1,83 @@
...
-  isLoading: boolean
-  hasError: boolean
+const goToExplorePage = () => {
+  /*
+    Below assumes that going from the details page url up one level will work,
+    for the current set of portals this assumption will hold true.
+  */
+  const lastLocation = window.location.href.split('/')
+  const lastPlace = lastLocation.slice(0, lastLocation.length - 1).join('/')
+  window.location.assign(lastPlace)
```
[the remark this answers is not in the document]
nickgros: Think this is safe, but the assumption made in the comment doesn't hold true anymore with the new DetailsPage tabs. This is only called in case of an error, so I don't think it's super urgent to fix.

src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -98,169 +115,73 @@ export default class DetailsPage extends React.Component<
...
-    */
-    const lastLocation = window.location.href.split('/')
-    const lastPlace = lastLocation.slice(0, lastLocation.length - 1).join('/')
-    window.location.assign(lastPlace)
-  }
+  const {
+    data: asyncJobStatus,
+    isLoading,
+    error: hasError,
+  } = SynapseQueries.useGetQueryResultBundleWithAsyncStatus(queryBundleRequest)
```
nickgros, line 125: Use react-query hook to fetch bundle

src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -98,169 +115,73 @@ export default class DetailsPage extends React.Component<
...
+          }
+        </div>
+      </div>
+    )
   } else {
-    console.error('Could not scroll to element with index ', index)
-  }
-}
-
-const SideNavMenu: React.FC<{
```
nickgros, line 208: Moved to its own module

src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -1,44 +1,82 @@
...
+  /*
+    Below assumes that going from the details page url up one level will work,
+    for the current set of portals this assumption will hold true.
+  */
+  const lastLocation = window.location.href.split('/')
+  const lastPlace = lastLocation.slice(0, lastLocation.length - 1).join('/')
+  window.location.assign(lastPlace)
+}
+
+function HeadlineWithLink(props: { title: string; id: string }) {
```
nickgros, line 47: Component that shows a "link" icon on hover to indicate that a section is link-able

src/portal-components/DetailsPage/DetailsPageTabs.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { DetailsPageTabProps } from 'types/portal-util-types'
+
+export type DetailsPageTabsProps = {
+  tabConfigs: DetailsPageTabProps[]
+  loading: boolean
+  showMenu: boolean
+  queryResultBundle?: QueryResultBundle
+}
+
+const DetailsPageTabs: React.FunctionComponent<DetailsPageTabsProps> = (
```
nickgros, line 18: Moved + refactored, so the GitHub difftool thinks it's brand new

src/portal-components/DetailsPage/SideNavMenu.tsx
```diff
@@ -0,0 +1,76 @@
...
+  window.history.pushState(null, '', `#${id}`)
+  const wrapper = document.querySelector<HTMLElement>(`#${id}`)
+  if (wrapper) {
+    scrollToWithOffset(wrapper)
+  } else {
+    console.error('Could not scroll to element with id ', id)
+  }
+}
+
+export const SideNavMenu: React.FC<{
```
nickgros, line 19: Moved from DetailsPage module

src/shared-config/sharedRoutes.ts
```diff
@@ -1,10 +1,23 @@
...
+  hideRouteFromNavbar: true,
+  synapseConfigArray: [
+    {
+      name: 'Header',
+      props: undefined,
+      isOutsideContainer: true,
+    },
+  ],
+}
+
```
nickgros, line 15: Deleted the home component, which just showed the Header component. Add it back as a (shared) route.

src/tests/RouterResolver.test.tsx
```diff
@@ -1,126 +0,0 @@
...
-
-  it('Finds a render-able child route if a fallback route has no synpaseConfig', () => {
-    // In this case, /Explore is NOT renderable so we need to find a child route that is
-    const [route, pathname] = getRouteFromParams('/Explore/Not-A-Valid-Route')
-    const exploreRoutes = routesConfig[EXPLORE_INDEX] as NestedRoute
-    const firstRoute = exploreRoutes.routes[0] as NestedRoute
-    expect(route).toEqual(firstRoute)
-    expect(pathname).toEqual('/Explore/Data')
-  })
-})
```
nickgros, line 68: We're losing this functionality. Is it important?

src/types/portal-config.ts
```diff
@@ -290,31 +295,32 @@ export type SynapseConfig = (
...
   target?: string | undefined
   link?: string
   icon?: string
-  synapseConfigArray?: SynapseConfigArray
 }
 
-export interface BaseRoute extends RouteOptions {
-  isNested: false
+export type ConfigRoute = RouteOptions & {
+  /** If true, will render the components specified in `synapseConfigArray` if on a child page. See react-router's exact */
```
nickgros, line 312:
```suggestion
  /** See react-router's exact */
```

src/AppInitializer.tsx
```diff
@@ -122,8 +122,6 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
   componentDidUpdate(prevProps: Props) {
     // https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md
```
jay-hodgson, line 123: Looks like comment link is related to scroll, so it can also probably also be removed.

src/AppInitializer.tsx
```diff
@@ -122,8 +122,6 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
...
     if (this.props.location !== prevProps.location) {
-      // scroll to the top
-      window.scrollTo(0, 0)
```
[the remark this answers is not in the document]
jay-hodgson: I think this is when following a link (to another page in the app) that was lower on a page.  Does this still behave as expected (where the new page top content is shown when loaded)?  If this breaks, then perhaps the latest suggested solution would avoid page jumping problems?  https://v5.reactrouter.com/web/guides/scroll-restoration

src/configurations/adknowledgeportal/routesConfig.ts
jay-hodgson: This explanation was very helpful

src/portal-components/DetailsPage/DetailsPage.tsx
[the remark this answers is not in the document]
jay-hodgson: I'm not sure if this alert is valid.  Is window.location.href considered a user-provided value?

src/portal-components/DetailsPage/DetailsPage.tsx
jay-hodgson: Nice!

src/portal-components/DetailsPage/DetailsPageTabs.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { url } = useRouteMatch()
+  const urlWithTrailingSlash = `${url}${url.endsWith('/') ? '' : '/'}`
+  const { search } = useLocation()
+  return (
+    <>
+      <RedirectWithQuery
+        exact={true}
+        from={urlWithTrailingSlash}
+        to={`${urlWithTrailingSlash}${tabConfigs[0].uriValue}`}
+      />
```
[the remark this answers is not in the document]
jay-hodgson: I think this addresses the concern that I mentioned earlier.  Thanks!

src/portal-components/StatefulButtonControlWrapper.tsx
```diff
@@ -1,11 +1,11 @@
...
-import { SynapseConfigArray } from 'types/portal-config'
-import { SynapseComponentWithContext } from '../SynapseComponentWithContext'
+import { SynapseConfig } from 'types/portal-config'
 import QueryCount from 'synapse-react-client/dist/containers/QueryCount'
 import { RouteControl, RouteControlProps } from 'RouteControl'
+import { SynapseComponent } from 'SynapseComponent'
 
 export type StatefulButtonControlConfigs = {
-  synapseConfigArray: SynapseConfigArray
+  synapseConfigArray: SynapseConfig[]
```
jay-hodgson, line 8: Thanks for removing SynapseConfigArray in this PR

src/tests/RouterResolver.test.tsx
jay-hodgson: Is there a use case for this scenario? Someone linked to a route that has no synapseConfigs, and we need to find a random subroute that has synapseConfigs?

jay-hodgson (approved):

src/configurations/adknowledgeportal/synapseConfigs/studies.ts
nickgros: The DetailsPageTabs component has a built-in redirect that should handle that particular case

src/tests/RouterResolver.test.tsx
nickgros: I think without it, we will probably need to add a 404 page. I can't think of any other real case

PR #804, opened 2022-07-14
PORTALS-2234: remove isomorphic fetch polyfill

jay-hodgson:


src/tests/setupTests.ts
```diff
@@ -1,4 +1,3 @@
-import 'isomorphic-fetch' // polyfill for fetch
```
nickgros, line 1: can we just `import 'whatwg-fetch'` instead? It's currently a dependency, and I'm actually going to use that in SRC in favor of the react-app-polyfill packages since they trigger Jest's memory leak detection.

jay-hodgson: I have no preference.  Updated

PR #808, opened 2022-07-20
PORTALS-2228 - "Copy link" and fix redirect

nickgros:
- Clicking the link copies the URL to the section instead of navigating to it
- Fix an issue where the DetailsPageTab redirect would break navigating between tabs.
- Fix ID mismatch between SideNavMenu and rendered sections

Clip of "Copy link" functionality:

https://user-images.githubusercontent.com/17580037/180071249-4cbf5bdd-241e-4845-b51c-041283b6fe6c.mov

### DIFF
--- a/src/portal-components/DetailsPage/DetailsPage.tsx
+++ b/src/portal-components/DetailsPage/DetailsPage.tsx
@@ -395,7 +395,7 @@ export const DetailsPageSynapseConfigArray: React.FC<{
   const synapseConfigContent = (
     <>
       {synapseConfigArray.map((el: RowSynapseConfig, index) => {
-        const id = getComponentId(el, '', index)
+        const id = getComponentId(el)
         const { resolveSynId, showTitleSeperator = true } = el
         const key = JSON.stringify(el)
         const hasTitleFromSynId = resolveSynId && resolveSynId.title
--- a/src/portal-components/DetailsPage/DetailsPage.tsx
+++ b/src/portal-components/DetailsPage/DetailsPage.tsx
@@ -21,6 +21,7 @@ import {
   QueryBundleRequest,
   QueryResultBundle,
 } from 'synapse-react-client/dist/utils/synapseTypes/'
+import Tooltip from 'synapse-react-client/dist/utils/tooltip/Tooltip'
 import { SynapseComponent } from 'SynapseComponent'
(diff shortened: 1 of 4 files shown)


src/portal-components/DetailsPage/SideNavMenu.tsx
```diff
@@ -63,7 +63,7 @@ export const SideNavMenu: React.FC<{
               onClick={
                 isDisabled
                   ? undefined
-                  : () => handleMenuClick(getComponentId(el))
+                  : () => handleMenuClick(getComponentId(el, '', index))
```
nickgros, line 66: I don't like that we're using the index for the ID. It will break links to sections if content is added/reorganized within the section. But without the index, we may have collisions. Maybe each section should require a unique, human-readable identifier?

src/portal-components/DetailsPage/DetailsPage.tsx
```diff
@@ -70,7 +68,26 @@ function HeadlineWithLink(props: { title: string; id: string }) {
...
-          <IconSvg options={{ icon: 'link', padding: 'left' }} />
+          <Tooltip
+            title={copied ? 'Copied' : 'Copy link to section'}
+            placement="right"
+          >
+            <div
+              style={{ cursor: 'pointer' }}
+              onClick={() => {
+                const urlWithoutHash = window.location.href.replace(
+                  window.location.hash,
```
jay-hodgson, line 79: Does this contain the '#' character?

jay-hodgson (approved):

src/portal-components/DetailsPage/DetailsPage.tsx
nickgros: Yes. Checked in Chrome, FF, Safari.

src/portal-components/DetailsPage/utils.ts
```diff
@@ -6,10 +6,19 @@ import { scrollToWithOffset } from 'utils'
 export function getComponentId(
   rowSynapseConfig: RowSynapseConfig,
   entityTitle: string = '',
-  index: number = 0,
```
nickgros, line 9: I decided to drop the index. If we have a collision, then we should update the config titles to not collide.

PR #813, opened 2022-08-09
PORTALS-2240

yjyoo773:
Ticket required to create tooltip for Dataset Tabs too. However, currently there is no tab for datasets.


yjyoo773:
Screen Shot 2022-08-09 at 3 36 56 PM
Screen Shot 2022-08-09 at 3 37 04 PM

jay-hodgson (approved):

PR #822, opened 2022-08-23
PORTALS-2289

jay-hodgson:
After fixes to query wrapper plot nav padding:
Screen Shot 2022-08-23 at 3 08 25 PM
Screen Shot 2022-08-23 at 3 08 44 PM


jay-hodgson:
Center Consortia:
Screen Shot 2022-08-23 at 3 19 51 PM

PR #824, opened 2022-08-25
Upgrade to React 18 and fix issues caused by upgrade

nickgros:
- Requires new version of SRC
- Refactored AppInitializer to be a functional component
	- I was getting the "Breaking Rules of Hooks" error, and I didn't have duplicate copies of React. Seemed to be caused by hooks called in wrappers. Issue fixed after refactor.


src/AppInitializer.tsx
```diff
@@ -31,217 +30,221 @@
...
+      const cookieValue = {
+        isInvokingDownloadTable,
+        callbackUrl: window.location.href,
+        logoUrl: icon,
+        portalName: name,
+      }
+      // expire after 10 seconds
+      const domainValue = window.location.hostname
+        .toLowerCase()
+        .includes('.synapse.org')
```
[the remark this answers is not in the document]
nickgros: Duplicates [#2](https://github.com/Sage-Bionetworks/portals/security/code-scanning/2)

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
 
 const COOKIE_CONFIG_KEY = 'org.sagebionetworks.security.cookies.portal.config'
 
-class AppInitializer extends React.Component<Props, AppInitializerState> {
```
nickgros, line 34: I have no idea why I was getting "Invalid use of hooks" before converting this to a functional component and using hooks instead of passing props with HOCs, but the error went away once I changed it.

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
...
-  constructor(props: Props) {
-    super(props)
-    this.state = {
-      token: undefined,
-      hasCalledGetSession: false,
-      userProfile: undefined,
-      showLoginDialog: false,
-      synapseRedirectUrl: undefined,
+/** On mount, update the document title and meta description using data from the portal config */
+function useSetDocumentMetadataFromConfig() {
```
nickgros, line 39: We were doing many different things on `componentDidMount`, so I tried to turn each into it's own hook/instance of `useEffect`

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
...
-    this.setState({
-      showLoginDialog: false,
-    })
-  }
+/**
+ * State and helpers for managing a user session in the portal
+ * @param setShowLoginDialog
+ * @returns
+ */
+function useSession(
```
nickgros, line 55: This could be cleaned up further, but would take some thought that I haven't given it yet

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
...
-      this.setState({
-        userProfile,
-      })
+      setUserProfile(userProfile)
     } catch (e) {
       console.error('Error on getSession: ', e)
       // intentionally calling sign out because there token could be stale so we want
       // the stored session to be cleared out.
-      this.initAnonymousUserState()
+      initAnonymousUserState()
```
nickgros, line 92: Would need to merge here with #828

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
...
-      const windowAny: any = window
-      const gtag = windowAny.gtag
-      if (gtag) {
-        gtag('config', 'G-CEKFPZDZX7', {
-          page_location: window.location.href,
-          page_path: `/${this.props.location.pathname}`,
-        })
-      }
-    }
-  }
```
nickgros, line 135: Moved to `useAnalytics.ts` and called here at line 128.

src/AppInitializer.tsx
```diff
@@ -31,217 +35,245 @@ export type SignInProps = {
...
+
+  const onSignInClicked = useCallback(() => {
+    setShowLoginDialog(true)
+  }, [])
+
+  const handleCloseLoginDialog = React.useCallback(() => {
+    setShowLoginDialog(false)
+  }, [])
+
+  const clonedChildren = useMemo(
```
nickgros, line 227: This clone is now memoized, which may improve performance

src/Footer.tsx
```diff
@@ -54,16 +54,28 @@ export const Footer: React.SFC<{}> = () => {
...
+          >
             About
           </a>
         )}
       </div>
     </footer>
   )
 }
+
+export default React.memo(Footer)
```
nickgros, line 81: Has no state, so we can tell React not to re-render it when a parent updates.

src/portal-components/Header.tsx
```diff
@@ -40,3 +40,5 @@ export const Header: React.SFC<{}> = () => {
     </header>
   )
 }
+
+export default React.memo(Header)
```
nickgros, line 44: Has no state, so we can tell React not to update it when a parent updates.

src/withLocation.tsx
```diff
@@ -1,12 +0,0 @@
-import React from 'react'
```
nickgros, line 1: This was only used by `AppInitializer`, which can now just call the `useLocation` hook itself

src/AppInitializer.tsx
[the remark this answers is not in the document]
jay-hodgson: Would changing includes() to endsWith() still function properly?

jay-hodgson (approved):

PR #828, opened 2022-08-30
PORTALS-2293

jay-hodgson:
@nickgros : Brought the sledgehammer instead of the scalpel for this proposed (quick) fix

### DIFF
--- a/src/AppInitializer.tsx
+++ b/src/AppInitializer.tsx
@@ -83,7 +83,11 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
       console.error('Error on getSession: ', e)
       // intentionally calling sign out because there token could be stale so we want
       // the stored session to be cleared out.
-      this.initAnonymousUserState()
+      SynapseClient.signOut(() => {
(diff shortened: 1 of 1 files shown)


src/AppInitializer.tsx
```diff
@@ -83,7 +83,11 @@ class AppInitializer extends React.Component<Props, AppInitializerState> {
       console.error('Error on getSession: ', e)
       // intentionally calling sign out because there token could be stale so we want
       // the stored session to be cleared out.
-      this.initAnonymousUserState()
+      SynapseClient.signOut(() => {
+        // PORTALS-2293: if the token was invalid (caused an error), reload the app to ensure all children
+        // are loading as the anonymous user
+        window.location.reload()
```
jay-hodgson, line 89: Only called when there was an error on getSession using the access token from the cookie

nickgros: If we think the issue is caused by react-query, we could blow away the entire query cache here

nickgros (approved):

PR #830, opened 2022-08-31
PORTALS-2293: when deployed, getAccessTokenFromCookie throws an error…

jay-hodgson:
… when anonymous


jay-hodgson:
Merging because staging site is currently down (when anonymous) due to this bug.

PR #835, opened 2022-09-13
Bump SRC, markdown-it, drop sanitize-html

nickgros:


jay-hodgson (approved):

jay-hodgson:
Is sanitize-html bundled with SRC?

nickgros:
> Is sanitize-html bundled with SRC?
Yes, see https://github.com/Sage-Bionetworks/Synapse-React-Client/commit/3867e7767455f0455821c028650acd8fa113d543

PR #837, opened 2022-09-14
PORTALS-2313

jay-hodgson:
init ARK portal config using other portal config/data


jay-hodgson:
dHealth also has custom icons, so leaving these in here as an example.
The ARKWelcomePage component is custom, and I think we can try to add all of the special layout and home page background images here.
The ARK icon currently has a very large mask embedded in it, but I've asked Adam for help exporting a usable asset from figma.

jay-hodgson:
This portal is a work-in-progress.  I'd like to deploy as soon as possible to verify the infra

PR #867, opened 2022-10-25
PORTALS-2157: Rename facetAliases to columnAliases

nickgros:
- Account for changes caused by Sage-Bionetworks/Synapse-React-Client#1854
- Requires upgrade to next version of SRC containing changes


jay-hodgson (approved):
Pending SRC update and a successful build

PR #875, opened 2022-10-31
IT-2157: Fix news site url

xschildw:
Quick fix...
@jay-hodgson , is that URL going to be a problem to display in the site itself (i.e. should it be https://news.arkportal.synapse.org)?


xschildw:
This is already done.

PR #899, opened 2022-11-17
Update to 3.1.13, and use palette from SRC

jay-hodgson:


src/configurations/nf/synapseConfigs/studies.ts
```diff
@@ -68,14 +68,14 @@ export const studyCardConfiguration: CardConfiguration = {
...
+        Available: { icon: 'data', sx: {fill: '#28A745'} },
+        'Partially Available': { icon: 'data', sx: {fill: '#DE9A1F' } },
+        'Under Embargo': { icon: 'dataLocked', sx: {fill: '#D46D1E' } },
+        None: { icon: 'data', sx: {fill: '#BBBBBC' }},
       },
       studyStatus: {
-        Active: { icon: 'reload', color: '#28A745' },
-        Completed: { icon: 'check', color: '#B2242A' },
+        Active: { icon: 'reload', sx: {fill: '#28A745' }},
+        Completed: { icon: 'check', sx: {fill: '#B2242A' }},
```
nickgros, line 78: Surprised you had to use fill instead of color for these. Not sure if it makes sense to pull these colors from the palette.

nickgros (approved):

src/configurations/nf/synapseConfigs/studies.ts
jay-hodgson: I think the dataStatus and studyStatus icons were removed, so I think these configurations could be deleted.  dataType is still used.

jay-hodgson: Changed to 'color' for the example configuration.

PR #905, opened 2022-11-30
PORTALS-2433: add other metadata for social media preview cards

jay-hodgson:


jay-hodgson:
Going to pull this in and take a look at the jenkins build and staging sites to see if this build process is working as expected.

PR #911, opened 2022-12-07
PORTALS-2402

jay-hodgson:


src/configurations/nf/routesConfig.ts
```diff
@@ -48,6 +48,12 @@ const routes: GenericRoute[] = [
     path: '',
     exact: true,
```
nickgros, line 49: Could probably put the component in a route with `path: ''` and `exact: false` if you wanted it to appear on any page in the portal.

src/portal-components/nf/NFSurveyToast.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
+import { useCookies } from 'react-cookie'
+// import { SynapseComponents } from 'synapse-react-client'
```
nickgros, line 3:
```suggestion
```

nickgros (approved):

src/configurations/nf/routesConfig.ts
jay-hodgson: Great idea!

nickgros (approved):

